Issue #21, opened 2018-09-15
Missing Plugin Exception: implementation not found for method readAll


jonsamwell:
Hi,

I am getting the above plugin exception running as an installed app (in release mode) on a Google Pixel XL running Android 9 SDK 28.  I haven't got another real device to test on at the moment but it works on the emulator.  

Unfortunately, that is all the information I have.  Let me know if you can think of any other things I can try or ways to get more information.

Update: I get the same result if I build and release the application from the flutter dev or beta channel.  It does however work if I run in release mode on the actual device during development.  Once it is install (via the play store) it does not work.

mogol:
Hi @jonsamwell 
Could you post full log ?

jonsamwell:
How would I get the flutter logs? This only happens when it is build in
release mode and installed via the play store. Is there a flutter log I can
pull off the phone somewhere? I have my application level log but that only
gives as much information as I gave.

mogol:
Do you have stacktrace for the exception?

Also could you check android code in your project? Does it contain implementation of readAll method?

It looks like there are different versions of plugin dart code and native code in your project. Somehow android project was built with 3.0 version, but dart code uses 3.1

hnvn:
I confirm this error when running on release mode. Here's the stacktrace:
```
 Registration failed
java.security.InvalidKeyException: Failed to unwrap key
   at android.security.keystore.AndroidKeyStoreCipherSpiBase.engineUnwrap(AndroidKeyStoreCipherSpiBase.java:682)
   at javax.crypto.Cipher.unwrap(Cipher.java:2435)
   at com.it_nomads.fluttersecurestorage.ciphers.RSACipher18Implementation.unwrap(RSACipher18Implementation.java:46)
   at com.it_nomads.fluttersecurestorage.ciphers.StorageCipher18Implementation.<init>(StorageCipher18Implementation.java:48)
   at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.<init>(FlutterSecureStoragePlugin.java:46)
   at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.registerWith(FlutterSecureStoragePlugin.java:33)
   at io.flutter.plugins.GeneratedPluginRegistrant.registerWith(GeneratedPluginRegistrant.java:39)
   at vn.com.bibomart.app.MainActivity.onCreate(MainActivity.java:17)
   at android.app.Activity.performCreate(Activity.java:6720)
[...]

Issue #13, opened 2018-05-22
3.1.0: Failed to unwrap key


jonasbark:
I did the upgrade to 3.1.0 from 3.0.0. Execution worked just fine, however after cleaning the app data and reinstalling the app this error showed up - which strikes me as odd.

```
E/FlutterSecureStoragePl(10643): Registration failed
E/FlutterSecureStoragePl(10643): java.security.InvalidKeyException: Failed to unwrap key
E/FlutterSecureStoragePl(10643): 	at android.security.keystore.AndroidKeyStoreCipherSpiBase.engineUnwrap(AndroidKeyStoreCipherSpiBase.java:682)
E/FlutterSecureStoragePl(10643): 	at javax.crypto.Cipher.unwrap(Cipher.java:2121)
E/FlutterSecureStoragePl(10643): 	at com.it_nomads.fluttersecurestorage.ciphers.RSACipher18Implementation.unwrap(RSACipher18Implementation.java:46)
E/FlutterSecureStoragePl(10643): 	at com.it_nomads.fluttersecurestorage.ciphers.StorageCipher18Implementation.<init>(StorageCipher18Implementation.java:48)
E/FlutterSecureStoragePl(10643): 	at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.<init>(FlutterSecureStoragePlugin.java:46)
E/FlutterSecureStoragePl(10643): 	at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.registerWith(FlutterSecureStoragePlugin.java:33)
E/FlutterSecureStoragePl(10643): 	at io.flutter.plugins.GeneratedPluginRegistrant.registerWith(GeneratedPluginRegistrant.java:33)
E/FlutterSecureStoragePl(10643): 	at de.ffuf.uffl2.MainActivity.onCreate(MainActivity.kt:11)
E/FlutterSecureStoragePl(10643): 	at android.app.Activity.performCreate(Activity.java:6998)
E/FlutterSecureStoragePl(10643): 	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1230)
E/FlutterSecureStoragePl(10643): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2899)
E/FlutterSecureStoragePl(10643): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3046)
[...]

PR #23, opened 2018-09-20
fix bug: initializing plugin failed by cannot unwrap existing key.

hnvn:
Related to bug #21 , that's a weird case when existing AES key cannot be unwrapped, so I add some codes to catch this error and re-create a new key in this case.

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
@@ -38,12 +38,14 @@ public class StorageCipher18Implementation implements StorageCipher {
         cipher = Cipher.getInstance("AES/CBC/PKCS7Padding");
 
         if (aesKey != null) {
+            byte[] encrypted;
             try {
-                byte[] encrypted = Base64.decode(aesKey, Base64.DEFAULT);
+                encrypted = Base64.decode(aesKey, Base64.DEFAULT);
                 secretKey = rsaCipher.unwrap(encrypted, KEY_ALGORITHM);
                 return;
             } catch (Exception e) {
                 Log.e("StorageCipher18Impl", "unwrap key failed", e);
+                encrypted = new byte[0];
             }
         }
 
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.util.Base64;
+import android.util.Log;
 
 import java.security.Key;
 import java.security.SecureRandom;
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
```diff
@@ -33,22 +34,26 @@ public StorageCipher18Implementation(Context context) throws Exception {
...
+                byte[] encrypted = Base64.decode(aesKey, Base64.DEFAULT);
+                secretKey = rsaCipher.unwrap(encrypted, KEY_ALGORITHM);
+                return;
+            } catch (Exception e) {
+                Log.e("StorageCipher18Impl", "unwrap key failed", e);
+            }
+        }
+
+        byte[] key = new byte[keySize];
+        secureRandom.nextBytes(key);
```
mogol: Doesn’t it create new key every time ?
Also shouldn’t be data deleted too?

hnvn: No, I don't. You maybe get me wrong. If the current key can be unwrapped, I use this key and I call `return` so the codes of creating a new key are ignored.

mogol: Yes, my fault, haven't noticed return :+1:

What's about the data which was encrypted with "broken" key?

hnvn: yep, I forgot to delete data of the wrong key

android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/StorageCipher18Implementation.java
```diff
@@ -33,22 +34,28 @@ public StorageCipher18Implementation(Context context) throws Exception {
...
+                encrypted = new byte[0];
+            }
+        }
+
+        byte[] key = new byte[keySize];
+        secureRandom.nextBytes(key);
+        secretKey = new SecretKeySpec(key, KEY_ALGORITHM);
+
+        byte[] encryptedKey = rsaCipher.wrap(secretKey);
+        editor.putString(AES_PREFERENCES_KEY, Base64.encodeToString(encryptedKey, Base64.DEFAULT));
```
mogol: It wouldn't be possible to decrypt old data, that was saved with previous (broken) key.

hnvn: It's true. But I don't know why this bug only happens when running on release mode. In release mode, although I uninstall the previous app and reinstall it, I still get a previous key from SharePreference and this key is broken. Very strange. And when I run it on another device, it works without this problem. So, my code temporarily gets around this problem by deleting that 'ghost' key and create a new.

mogol: Just curious, 
could you try https://github.com/mogol/flutter_secure_storage/issues/13#issuecomment-421083742 this? would it help you?

PR #33, opened 2018-12-29
Added method contains

mirandawork:
Added method contains to check if a key exists.

Signed-off-by: André R. de Miranda <andre@miranda.work>

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
@@ -59,6 +59,13 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler {
                     result.success(null);
                     break;
                 }
+                case "contains": {
+                    String key = getKeyFromCall(call);
+
+                    boolean value = contains(key);
+                    result.success(value);
+                    break;
+                }
                 case "read": {
                     String key = getKeyFromCall(call);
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -59,6 +59,13 @@ public void onMethodCall(MethodCall call, Result result) {
                     result.success(null);
                     break;
                 }
+                case "contains": {
```
mogol: Could you please add implementation for iOS plugin?

lib/flutter_secure_storage.dart
```diff
@@ -17,6 +17,12 @@ class FlutterSecureStorage {
     return value;
   }
 
+  Future<bool> contains({@required String key}) async {
+    final bool value =
+    await _channel.invokeMethod('contains', <String, String>{'key': key});
```
mogol: What's about reusing "read" method and comparing with `null`?

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
mirandawork: Sorry. I don't have the development environment for IOS.

lib/flutter_secure_storage.dart
mirandawork: Using the 'read' method, if the key exists, is decrypted unnecessarily.

iassal:
@mogol I think we can close this since https://github.com/mogol/flutter_secure_storage/pull/139 got merged

PR #37, opened 2019-01-26
Replace Support with AndroidX

I-am-original:
Hi there. Here is update to use AndroidX instead Support lib. 
Please take a look and merge if it is ok for you. 

Cheers

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.1'
+        classpath 'com.android.tools.build:gradle:3.3.0'
     }
 }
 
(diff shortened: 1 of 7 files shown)


android/build.gradle
```diff
@@ -22,15 +22,14 @@ rootProject.allprojects {
...
     defaultConfig {
         minSdkVersion 18
-        targetSdkVersion 27
-        versionCode 5
-        versionName "3.1.1"
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        targetSdkVersion 28
+        versionCode 6
+        versionName "3.1.4"
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
mogol: There are no android tests, 
could `testInstrumentationRunner` be deleted?

I-am-original: @mogol as it was included before - I left it as it is. Maybe you would like to write some test ) at some point of time )

mogol: Does the plugin need any references to SupportLibrary/AndroidX?
Looks like if I remove any references to tests libs, then SupportLibrary/AndroidX are not needed.

I-am-original: @mogol  ok you can left it as it is. up to you

PR #45, opened 2019-03-21
add testing mock

Goxiaoy:
add mock for testing purpose

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,11 @@ android {
 ```
 *Note* By default Android backups data on Google Drive. It can cause exception java.security.InvalidKeyException:Failed to unwrap key. You need to disable backup to fix this issue. https://github.com/mogol/flutter_secure_storage/issues/13#issuecomment-421083742
 
+
+### Testing
+
+```
+//set initial values here
+FlutterSecureStorage.setMockInitialValues({})
+
+```
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -26,4 +26,26 @@ class FlutterSecureStorage {
   }
 
   Future<void> deleteAll() => _channel.invokeMethod('deleteAll');
+
+
+  @visibleForTesting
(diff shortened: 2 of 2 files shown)


mogol:
Hi @Goxiaoy 
Why is mock implementation needed? doesn't [Mockito](https://pub.dartlang.org/packages/mockito) cover this case?

Goxiaoy:
Because when I try to do unittest within codes rely on this package，error comes with no implementation on this platform, which means on my computer

Goxiaoy:
And it is learned from package shared_preferences

mogol:
Would it work in your case?

```dart

import 'package:flutter_secure_storage/flutter_secure_storage.dart';
import 'package:mockito/mockito.dart';
import 'package:test/test.dart';

void main() {
  test('Sample mock', () async {
    final storage = MockStorage();
    when(storage.read(key: 'name')).thenAnswer((_) => Future.value('Stub'));

    final name = await storage.read(key: 'name');
    expect(name, 'Stub');
  });
}

class MockStorage extends Mock implements FlutterSecureStorage{}
```

Goxiaoy:
It works for me. I am using dependency injection in my flutter project. so it's quite easy for me to replace FlutterSecureStorage to the mock class. But for those who did not, it's hard to replace it.

It's fine to close this PR. Thank you so much for your introduce of mockito.

tirth-stasis:
> Would it work in your case?
> 
> ```dart
> import 'package:flutter_secure_storage/flutter_secure_storage.dart';
> import 'package:mockito/mockito.dart';
> import 'package:test/test.dart';
> 
> void main() {
>   test('Sample mock', () async {
>     final storage = MockStorage();
>     when(storage.read(key: 'name')).thenAnswer((_) => Future.value('Stub'));
> 
>     final name = await storage.read(key: 'name');
>     expect(name, 'Stub');
>   });
> }
> 
> class MockStorage extends Mock implements FlutterSecureStorage{}
> ```

How to do that for `write` method?

PR #56, opened 2019-05-26
Added groupId for iOS keychain sharing

Maleandr:


mogol:
Hi @Maleandr 
First of all, thanks for the contribution! 

The only concern I have is how platform-specific API should be organised, as there could more different options.
E.g. one option is to rename `groupId` to `iOSGroupId` to highlight the difference.
```
Future<String> read({@required String key, String groupId})
```
vs
```
Future<String> read({@required String key, String iOSGroupId})
```
Another option can be
```
Future<String> read({@required String key, iOSOptions ios, AndroidOptions android})
```

Maleandr:
Separate system options looks nice. I will refactor accordingly.

lib/flutter_secure_storage.dart
```diff
@@ -12,40 +13,78 @@ class FlutterSecureStorage {
...
+  Future<void> deleteAll({iOSOptions iOptions, AndroidOptions aOptions}) =>
+      _channel.invokeMethod('deleteAll', <String, dynamic>{'options': _selectOptions(iOptions, aOptions)});
+
+  /// Select correct options based on current platform
+  Map<String, String> _selectOptions(iOSOptions iOptions, AndroidOptions aOptions) {
+    return Platform.isIOS ? iOptions?.params : aOptions?.params;
+  }
+}
+
+class Options {
```
mogol: Can it be `abstract class Options`? 

```
abstract class Options {
  Map<String, String> get params => _toMap();
  Map<String, String> _toMap();
}
```

Maleandr: Yep

mogol:
Thanks!

PR #66, opened 2019-07-25
Fix #58; fix No certificate found under alias

### LINKED ISSUE #58: No certificate found under alias error
`Caused by java.lang.Exception: No certificate found under alias: com.*.*.FlutterSecureStoragePluginKey
       at com.*.*.encryption.RSACipher18Implementation.getPublicKey + 91(RSACipher18Implementation.java:91)
       at com.*.*.encryption.RSACipher18Implementation.wrap + 38(RSACipher18Implementation.java:38)
       at com.*.*.encryption.StorageCipher18Implementation.<init> + 56(StorageCipher18Implementation.java:56)`

Hi, I've pulled this library into our native app ahead of launching our Flutter app and seen crash reports on a variety of Samsung Galaxy devices. Any idea what could be causing this issue?

prprhyt:
Fix  #58.
https://github.com/mogol/flutter_secure_storage/blob/6f57918a38af01f963245351ef0d8b5aa9272b6a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java#L113

I had this issue in pixel3a(Android 9.0).
But, after I run fixed code once, #58 is no longer reproduced in the original code.
So, when an alias has never been created, `ks.getKey(KEY_ALIAS, null)` is expected to return NULL, but it maybe not.

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
@@ -110,8 +110,16 @@ class RSACipher18Implementation {
         KeyStore ks = KeyStore.getInstance(KEYSTORE_PROVIDER_ANDROID);
         ks.load(null);
 
-        Key privateKey = ks.getKey(KEY_ALIAS, null);
-        if (privateKey == null) {
+        // see:
+        // https://stackoverflow.com/questions/36488219/android-security-keystoreexception-invalid-key-blob
+        KeyStore.Entry entry = null;
+        try {
+            entry = ks.getEntry(KEY_ALIAS, null);
+        } catch (Exception e) {
+            e.printStackTrace();
+            ks.deleteEntry(KEY_ALIAS);
+        }
+        if (entry == null) {
             createKeys(context);
         }
     }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
```diff
@@ -110,8 +110,16 @@ private void createRSAKeysIfNeeded(Context context) throws Exception {
         KeyStore ks = KeyStore.getInstance(KEYSTORE_PROVIDER_ANDROID);
         ks.load(null);
 
-        Key privateKey = ks.getKey(KEY_ALIAS, null);
-        if (privateKey == null) {
+        // see:
+        // https://stackoverflow.com/questions/36488219/android-security-keystoreexception-invalid-key-blob
+        KeyStore.Entry entry = null;
+        try {
+            entry = ks.getEntry(KEY_ALIAS, null);
```
mogol: Hi @prprhyt,
thanks for MR.

I have 2 questions:
* according to the SO link, why don't you add retry?
* if we delete KEY_ALIAS, it means we are not able to decode existing data from prefs. correct?

tuarrep:
This bug bubbled for me today. Cleaning app data fixed it

Issue #79, opened 2019-09-17
android mainfest file sync error


yuanzhiying:
ERROR: The minSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Remove minSdkVersion and sync project
Affected Modules: flutter_secure_storage

when i use this package in android project, sync error. The minSdkVersion should be declared in build.gradle file.

Issue #77, opened 2019-09-14
Erro android studio


RenanDelfanti:
ERROR: The minSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Remove minSdkVersion and sync project
Affected Modules: flutter_secure_storage

PR #80, opened 2019-09-17
Remove unnecessary manifest change

nate-eisner:
Since its pointed out [here](https://github.com/mogol/flutter_secure_storage/blob/bb86847add89f93b0ca03ddff3a4ae8f06bda498/android/build.gradle#L29). It is not necessary to have it in the manifest. Also, seems to be breaking builds having it there.

Fixes https://github.com/mogol/flutter_secure_storage/issues/79 & https://github.com/mogol/flutter_secure_storage/issues/77


mogol:
Hi @nate-eisner
Thanks, will check today

PR #91, opened 2019-11-28
Add macOS plugin.

schmidp:
I added a first version of a macOS plugin.
Basically its a copy of the iOS plugin with the smallest adoptions for macOS.
It would probably make sense to reuse the same code for macOS, but I wanted to share something now.


mogol:
Thanks!
I will take a look

nailgilaziev:
hello! what is a status of this PR?

lib/flutter_secure_storage.dart
```diff
@@ -71,10 +71,10 @@ abstract class Options {
   }
 }
 
-class iOSOptions extends Options {
-  final String _groupId;
+class IOSOptions extends Options {
```
spkersten, line 74: This is a backwards incompatible change. Probably better to revert that and do it separately.

mogol:
Linked to #106

maxisme:
Hey @schmidp  I tried adding:

`NSUserNotificationAlertStyle`: `alert` to my Resources/info.plist but my notifications are still acting as banners. Any idea why?

Thanks

esiqveland:
Any new thoughts on this?

mpfaff:
I've noticed that upon starting my app, I need to enter my password 3-5 times to unlock my keychain, and each time I'm selecting "always allow"

maxisme:
@mpfaff that is probably because you call the request to get the key from keychain 3-5 based on different keys? I store everything under one key in JSON to avoid this.

Roms1383:
Yeah @mpfaff I got a POC working with go-flutter on MacOS (using go-keychain under the hood) and the reason is that, in my implementation, each key represents a different account as @maxisme suggested. What is weird though is that whenever creating a new key/value in the keychain, the flutter app is supposed to be the owner of it, hence not requiring the user to enter his keychain password to grant the app access to it (as from Apple documentation). Even weirder, it seemed during development that every time I made changes to the code and recompile, I had to re-enter my keychain password for every single key/value entry (no matter selecting "always allow") : I'm missing something out of the picture, like maybe rebundling code changes makes my app look like a different app to the OS ? By the way the policy I use when I manipulate the entries is kSecAttrAccessibleAfterFirstUnlock.

Roms1383:
In case it helps other beginners like me :sweat_smile: I got some feedback from Flutter/Swift developers who actually confirm that, since I just develop locally as-is (using self-signed certificate under the hood probably), rebuilding the app makes it appear as a different app to the OS, hence why it asks the password after each code changes/recompiling. Solution for this is to sign the app with an Apple generated certificate.

maxisme:
Thank you for that solution @Roms1383

juliansteenbakker:
@mogol We can close this one since its code is very similar to #181 but with some breaking changes.

PR #95, opened 2019-12-03
Remove the entry when call 'write' with a null for 'value' param

ewertonrp:
Remove the entry when call **write** function with a _null_ for **value** param, instead of throw an exception on Android and do nothing on iOS.

This way the plugin will have the same behavior for both platforms, and will be cohesive with the return of the **read** function for a given key that not exists, which is null.

### DIFF
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -12,12 +12,15 @@ class FlutterSecureStorage {
   /// Encrypts and saves the [key] with the given [value].
   ///
   /// If the key was already in the storage, its associated value is changed.
-  /// [value] and [key] shoudn't be null.
+  /// If the value is null, deletes associated value for the given [key].
+  /// [key] shoudn't be null.
+  /// [value] required value
   /// [iOptions] optional iOS options
   /// [aOptions] optional Android options
   /// Can throw a [PlatformException].
-  Future<void> write({@required String key, @required String value, IOSOptions iOptions, AndroidOptions aOptions}) async =>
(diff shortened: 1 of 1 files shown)


lib/flutter_secure_storage.dart
```diff
@@ -12,12 +12,15 @@ class FlutterSecureStorage {
...
-  /// [value] and [key] shoudn't be null.
+  /// If the value is null, deletes associated value for the given [key].
+  /// [key] shoudn't be null.
+  /// [value] required value
   /// [iOptions] optional iOS options
   /// [aOptions] optional Android options
   /// Can throw a [PlatformException].
-  Future<void> write({@required String key, @required String value, IOSOptions iOptions, AndroidOptions aOptions}) async =>
-      _channel.invokeMethod('write', <String, dynamic>{'key': key, 'value': value, 'options': _selectOptions(iOptions, aOptions)});
+  Future<void> write({@required String key, @required String value, IOSOptions iOptions, AndroidOptions aOptions}) => value != null
```
mogol: Shouldn't it be just assert?

ewertonrp: Hi, 

As a plugin, was expected to have the same behaviour between platforms. So, we have two options:
- use assert to prevent pass null at write function;
- infer the removal of the value;

I think that the second one, infer the removal of the value, is more convenient because of the behaviour of the `read` function:
> Decrypts and returns the value for the given [key] or null if [key] is not in the storage.
 
Once the `read` function returns `null` if key is not in the storage, this allows us to infer that the key value will be removed if we write `null`.

Issue #30, opened 2018-11-04
broken build in latest flutter


Rockvole:
flutter_secure_storage: 3.1.2

Latest flutter build on ios gives :

	flutter build ios
	Building com.rockvole.kbmlViewer for device (ios-release)...
	Automatically signing iOS for device deployment using specified development team
	in Xcode project: SQ4ZRAJ4V2
	Running pod install...                                       3.1s
	Starting Xcode build...                                          
	Xcode build done.                                            5.8s
	Failed to build iOS app
	Error output from Xcode build:
	↳
		** BUILD FAILED **

	Xcode's output:
	↳
		/Users/david/.pub-cache/hosted/pub.dartlang.org/flutter_secure_storage-3.1.2
		/ios/Classes/FlutterSecureStoragePlugin.m:87:49: warning: values of type
		'OSStatus' should not be used as format arguments; add an explicit cast to
		'int' instead [-Wformat]
					NSLog(@"SecItemUpdate status = %d", status);
												   ~~   ^~~~~~
												   %d   (int)
		/Users/david/.pub-cache/hosted/pub.dartlang.org/flutter_secure_storage-3.1.2
		/ios/Classes/FlutterSecureStoragePlugin.m:95:46: warning: values of type
		'OSStatus' should not be used as format arguments; add an explicit cast to
		'int' instead [-Wformat]
					NSLog(@"SecItemAdd status = %d", status);
												~~   ^~~~~~
												%d   (int)
		3 warnings generated.
		<module-includes>:1:1: warning: umbrella header for module 'Flutter' does
		not include header 'FlutterNavigationController.h' [-Wincomplete-umbrella]
		#import "Headers/Flutter.h"
		^
		1 warning generated.
		1 warning generated.

	Encountered error while building for device.

Rockvole:
Never mind - it resolved after updating another plugin.

uscagr:
Hey, we are having the same issue on bitrise, what packages you updated for a succesful build?

khalithartmann:
would like to know too ... did you resolve ?

sotoyjuan:
Same here. What was the fix.

PR #101, opened 2020-01-08
Fix build warning. Issue #30

jthlim:
"warning: values of type 'OSStatus' should not be used as format arguments; add an explicit cast to 'int' instead"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [3.3.1+2]
+* Fix iOS build warning [Issue 30](https://github.com/mogol/flutter_secure_storage/issues/30)
+
 ## [3.3.1+1]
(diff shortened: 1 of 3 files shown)


mehmetf:
@mogol Please take a look.

We use this plugin internally at Google. I am the tech lead for supporting internal Google apps using Flutter. It would be great if you could add me (mehmetf@) as a collaborator to this repo so I can help review PRs and push them through. I am quite active on the core Flutter plugins (and firebase) repo.

jthlim:
> Please update changelog.md and pubspec.yaml as well.

Done

mehmetf:
I will merge this change since it is consistent with what this repo has been doing but for further reference, semver for Dart states that once a package is above 0.x, each minor increment should be explicit rather than using `+`. I.e. instead of `3.1.1+1`, it should be `3.1.2`.

Issue #50, opened 2019-04-09
iOS: Allow customizing kSecAttrAccessible


hpoul:
I store tokens in the keychain which I need to access from a background process (background_fetch, and `content-available` push notifications) - to do this items have to be stored with `kSecAttrAccessible` = `kSecAttrAccessibleAlways`.

I think, the current implementation is a bit inconsistent, because on Android the default is that items can always be accessed, while on iOS the secure_storage can only be accessed while the user has unlocked the device.

andersunloc:
This is really required for a lot of apps on the iOS side, and also the bug is pretty hard to test, so probably a lot of people will have issues when their data suddenly is not readable

fvisticot:
I have the same issue. I need to get information from the keychain when the app is in background.
Is there any solution / workarround ?

Issue #94, opened 2019-12-01
[iOS] Does not work when keychain accessed when app is in background


fvisticot:
I need to access the keychain on iOS to get the token to send HTTP request when app is in background.

Seems the same issue to: https://github.com/mogol/flutter_secure_storage/issues/50

fvisticot:
any update ?

mogol:
Hi @fvisticot 
So far no updates, but I will try my best to update the plugin over the weekends

mogol:
The feature was added in v3.3.2

PR #108, opened 2020-03-25
Support for specifying keychain accessibility options

negator:
Adds support for specifying [Keychain Item Accessibility](https://developer.apple.com/documentation/security/keychain_services/keychain_items/restricting_keychain_item_accessibility?language=objc)

The options are specified as an enum value passed in the IOSOptions object as parameter to the `write` function.

This allows us to be able to fetch secure values while the app is backgrounded, by specifying [`first_unlock`](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlock?language=objc) or [`first_unlock_this_device`](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly?language=objc). The default if not specified is [`unlocked`](https://developer.apple.com/documentation/security/ksecattraccessiblewhenunlocked?language=objc).

An example:
```
final options = IOSOptions(accessibility: IOSAccessibility.first_unlock);
await storage.write(key: key, value: value, iOptions: options);
```

resolves:
https://github.com/mogol/flutter_secure_storage/issues/50
https://github.com/mogol/flutter_secure_storage/issues/94

### DIFF
--- a/ios/Classes/FlutterSecureStoragePlugin.m
+++ b/ios/Classes/FlutterSecureStoragePlugin.m
@@ -47,12 +47,13 @@ static NSString *const InvalidParameters = @"Invalid parameter's type";
         NSString *key = arguments[@"key"];
         NSString *value = arguments[@"value"];
         NSString *groupId = options[@"groupId"];
+        NSString *accessibility = options[@"accessibility"];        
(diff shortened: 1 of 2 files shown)


mogol:
Thanks :+1:

Issue #116, opened 2020-04-14
Initial key generation hangs UI rendering


morrica:
The initial key generation on first run will hang UI rendering including initial Widget paints or animations. This is a big issue for a live app because a new user's first impression will be that the app is clunky and slow to load.

This appears to only affect running on real devices, perhaps because the emulator is able to use underlying PC hardware generate the key faster than is noticeable.

As is, this issue prevents the app from loading any UI for between 5 and 30 seconds on first launch on a Pixel 3 running latest Android 10 build QQ2A.200305.002. 

Modifying `FlutterSecureStoragePlugin.java` to wait for an actual channel method call to initialize the `StorageCipher` will result in animations completely hanging for a similar period of time.

An example app that demonstrates the problem is included below. To reproduce the issue, run the given app on a real device and time the period before the `LoadingScreen` is rendered. Then time it a second time after the `StorageCipher` has been initialized. To retest the first run case you can clear the app's storage on the test device. 

[test_app.zip](https://github.com/mogol/flutter_secure_storage/files/4477557/test_app.zip)

gargakshit:
I encountered the same issue yesterday, and it seems like either we have to use isolate or manually delay the key generation (workaround)

morrica:
@gargakshit, this issue has been fixed in the merged commit above. The fix has not yet been pushed to pub.dev, but you can depend on github code directly in your pubspec.yaml if necessary.

gargakshit:
Cool, I'll use the git version. Thanks for the timely response

PR #117, opened 2020-04-14
Fix for issue #116: Android hanging UI on StorageCipher initialization

morrica:
Fix for issue #116: Android hanging UI on StorageCipher initialization. 
Similar issue may or may not affect iOS.
Basic architecture example taken from https://github.com/flutter/flutter/issues/22024#issuecomment-479849362

This fixes the issue by deferring StorageCipher initialization until a channel method is invoked, and completing the initialization and channel method in a separate Thread that returns the result through an asynchronous wrapper.

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
@@ -3,6 +3,8 @@ package com.it_nomads.fluttersecurestorage;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.os.Handler;
+import android.os.Looper;
 import android.util.Base64;
 import android.util.Log;
 import androidx.annotation.NonNull;
@@ -31,8 +33,10 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
     private StorageCipher storageCipher;
     private static final String ELEMENT_PREFERENCES_KEY_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIHNlY3VyZSBzdG9yYWdlCg";
     private static final String SHARED_PREFERENCES_NAME = "FlutterSecureStorage";
+    private static Context applicationContext;  //Necessary for deferred initialization of storageCipher
 
     public static void registerWith(Registrar registrar) {
+      applicationContext = registrar.context().getApplicationContext();
       FlutterSecureStoragePlugin instance = new FlutterSecureStoragePlugin();
       instance.initInstance(registrar.messenger(), registrar.context());
     }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -64,50 +84,9 @@ public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
...
-                    result.notImplemented();
-                    break;
-            }
-
-        } catch (Exception e) {
-            result.error("Exception encountered", call.method, e);
-        }
+    public void onMethodCall(@NonNull MethodCall call, @NonNull Result rawResult) {
+        MethodResultWrapper result = new MethodResultWrapper(rawResult);
+        new Thread(new MethodRunner(call, result)).start();
```
mogol, line 89: How heavy is the creation of a thread for each call? Are there any risks?

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -52,6 +55,23 @@ public void initInstance(BinaryMessenger messenger, Context context) {
       }
     }
 
+    /**
+     * This must be run in a separate Thread from an async method to avoid hanging UI thread on
+     * live devices in release mode.
+     * The most convenient place for that appears to be onMethodCall().
+     */
+    private void ensureInitStorageCipher() {
+        if(storageCipher == null) {
```
mogol, line 64: it seems possible to bump into race condition here. Any sync primitives?

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -3,6 +3,8 @@
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.os.Handler;
+import android.os.Looper;
 import android.util.Base64;
 import android.util.Log;
 import androidx.annotation.NonNull;
```
mogol, line 10: Please remove androidx imports
It breaks non migrated apps

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
morrica: I think it would depend on your app workload. You could use an ExecutorService to keep a thread in a pool for you if it will be called often. If the secure storage methods were called repeatedly, like in a benchmarking scenario, a thread pool would probably perform much better. However, my app is light on secure storage calls so holding a thread that is rarely used in the pool probably wouldn't be a net gain.

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
morrica: I considered that. As written the code is not thread safe, but the channel methods are all run on the main UI thread in a serialized fashion (hence the UI blocking issue) so I don't *THINK* thread safety is an issue. Perhaps it would be better to add `synchronized` to the `ensureInitStorageCipher()` method signature to be 100% sure though.

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
morrica: Confused... This pull request didn't add androidx. Shouldn't it go away when merged?

mogol: 👍 you are right!

Issue #131, opened 2020-06-01
RSACipher fails to generate keys when phone locale is Arabic in android


hmawla:
Spent a while surfing the code in android and apparently KeyPairGenerator fails to generateKeyPair() when the phone is set to Arabic locale and I guess to any RTL locales.
This is due to a bug in google's own code in KeyPairGenerator when parsing the expiry DateTime according to the used locale.

I still haven't figure a way out of fixing this but will send a PR when I do.

Reference: [Here](https://issuetracker.google.com/issues/37095309)
Possible solution: [Here](https://stackoverflow.com/questions/35987510/illegalargumentexception-unparseable-date-during-androidkeystore-key-generatio/46602170#46602170)

hmawla:
Created a workaround and made a PR
https://github.com/mogol/flutter_secure_storage/pull/132

Omrankabalan:
Any updates ?

juliansteenbakker:
A fix has been merged in #132.

PR #132, opened 2020-06-01
Fixed crash when generating keys in android with RTL locales

hmawla:
Created a workaround according to the community that would switch the locale to English before generating the keys, then switches back to the default one after that.

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
@@ -1,22 +1,24 @@
 package com.it_nomads.fluttersecurestorage.ciphers;
 
-import android.annotation.SuppressLint;
 import android.content.Context;
+import android.content.res.Configuration;
+import android.content.res.Resources;
 import android.os.Build;
 import android.security.keystore.KeyGenParameterSpec;
 import android.security.keystore.KeyProperties;
 import android.security.keystore.StrongBoxUnavailableException;
+import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


hmawla:
#131

mogol:
@hmawla good catch 👍 
Let me test and merge after.

iassal:
@hmawla awesome stuff! 

@mogol when can this be released?

android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
```diff
@@ -168,6 +186,6 @@ private void createKeys(Context context) throws Exception {
             kpGenerator.initialize(spec);
             kpGenerator.generateKeyPair();
         }
+        setLocale(localeBeforeFakingEnglishLocale);
```
iassal, line 189: @mogol Just wondering if this should be encapsulated in a finally block instead? Would this have any side effects if an exception was thrown and this wasn't called?

hmawla:
@iassal you can use my git if you need it urgently by adding [this ](https://pastebin.com/7TTcRLES)to your pubspec.yaml file
I won't be updating it furthermore until the next release is published on pub.dev.
Though make sure to follow @mogol when the next release is ready to switch back.

android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
hmawla: yes, it makes sense to do so just in case...

PR #133, opened 2020-06-05
Make MethodResultWrapper static and methodResult final

wytesk133:
MethodResultWrapper does not reference the enclosing class and, therefore, can be static.

methodResult is only assigned during initialization and, therefore, can be final.


mogol:
thanks @wytesk133

PR #134, opened 2020-06-05
Return the error as String rather than Exception

wytesk133:
StandardMessageCodec does not accept Exception values and, therefore, will throw an IllegalArgumentException.

https://github.com/flutter/engine/blob/7f388bf814d302f54379f8528ad2366aa03d5e28/shell/platform/android/io/flutter/plugin/common/StandardMessageCodec.java#L208-L280


mogol:
Thanks @wytesk133

wytesk133:
Actually, I think this PR can be reverted.

Support for Throwable was added in https://github.com/flutter/engine/pull/18490

PR #135, opened 2020-06-05
Fix Double-Checked Locking using volatile

wytesk133:
https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html


mogol:
@wytesk133 could you please elaborate why you think synchronization is not required there?

android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -59,18 +59,14 @@ public void initInstance(BinaryMessenger messenger, Context context) {
      * live devices in release mode.
      * The most convenient place for that appears to be onMethodCall().
      */
-    private void ensureInitStorageCipher() {
-        if(storageCipher == null) { //Check to avoid unnecessary entry into syncronized block
```
mogol, line 63: @wytesk133 according to the article, wouldn't making storageCipher `volatile` be better fix? 
Without synchronization it's seems race condition could break entire storage due to different aes keys.

Issue #52, opened 2019-04-20
App Crashes on Startup (Android 4.4)


RuckyZucky:
App Crashes on Samsung Galaxy S3 Neo (Android 4.4).
Exception in Logcat:
```04-20 16:30:42.742 3794-3794/com.it_nomads.fluttersecurestorageexample E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.it_nomads.fluttersecurestorageexample, PID: 3794
    java.lang.VerifyError: com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation
        at com.it_nomads.fluttersecurestorage.ciphers.StorageCipher18Implementation.<init>(StorageCipher18Implementation.java:31)
        at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.<init>(FlutterSecureStoragePlugin.java:46)
        at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin.registerWith(FlutterSecureStoragePlugin.java:33)
        at io.flutter.plugins.GeneratedPluginRegistrant.registerWith(GeneratedPluginRegistrant.java:14)
        at com.it_nomads.fluttersecurestorageexample.MainActivity.onCreate(MainActivity.java:12)
        at android.app.Activity.performCreate(Activity.java:5451)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1093)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2292)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2386)
        at android.app.ActivityThread.access$900(ActivityThread.java:169)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1277)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:136)
        at android.app.ActivityThread.main(ActivityThread.java:5476)
        at java.lang.reflect.Method.invokeNative(Native Method)
        at java.lang.reflect.Method.invoke(Method.java:515)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1283)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1099)
        at dalvik.system.NativeStart.main(Native Method)
```

[...]

PR #151, opened 2020-07-29
Fix RSACipher18Implementation bug on startup at API level 19 and lower

ewertonrp:
Fix app crash on Startup at Android 4.3 (api 18) and 4.4 (api 19).

The api level conditions are all correct, but because of the suppress lint annotation, the catch block with StrongBoxUnavailableException alert was suppressed and that code causes some kind of build optimization bug.

This fix the issue #52

### DIFF
--- a/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
+++ b/android/src/main/java/com/it_nomads/fluttersecurestorage/ciphers/RSACipher18Implementation.java
@@ -132,7 +132,6 @@ class RSACipher18Implementation {
         resources.updateConfiguration(config, resources.getDisplayMetrics());
     }
 
-    @SuppressLint("NewApi")
     private void createKeys(Context context) throws Exception {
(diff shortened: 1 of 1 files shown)


ewertonrp:
You're right. I generated a version from this branch for internal test only and the result was as follows:
- We had no more fatal errors, closing the application at startup;
- However, on some devices, we are having errors when using the plugin;

The correct way will be to identify why on some devices it is not starting correctly.

If I get some free time, I help investigate.

Thank you

juliansteenbakker:
I am closing this since there already is a check on the API SDK right now:
```dart
           if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
                spec = makeAlgorithmParameterSpecLegacy(context, start, end);
            } else {
                KeyGenParameterSpec.Builder builder = new KeyGenParameterSpec.Builder(KEY_ALIAS, KeyProperties.PURPOSE_DECRYPT | KeyProperties.PURPOSE_ENCRYPT)

```

PR #160, opened 2020-09-22
Remove unused dart:async import

MichaelRFairhurst:
This change does require depending on Dart >=2.1, if that's acceptable.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -18,4 +18,4 @@ dependencies:
     sdk: flutter
 
 environment:
-    sdk: ">=1.19.0 <3.0.0"
+    sdk: ">=2.1.0 <3.0.0"
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -1,4 +1,3 @@
-import 'dart:async';
 import 'dart:io';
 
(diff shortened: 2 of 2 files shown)


MichaelRFairhurst:
ping @mogol

MichaelRFairhurst:
ping again @mogol

mogol:
Hey @MichaelRFairhurst 
thanks for MRs, looking into

ps Need to check notifications, I missed the notifications again

pubspec.yaml
```diff
@@ -18,4 +18,4 @@ dependencies:
     sdk: flutter
 
 environment:
-    sdk: ">=1.19.0 <3.0.0"
+    sdk: ">=2.1.0 <3.0.0"
```
mogol, line 21: Why do we need this?

MichaelRFairhurst: As of Dart 2.1, Future and Stream have been exported from dart:core. I'm adding tooling to catch these imports, and fixing them where I can.

Dart 2.1 was launched Nov 15, 2018, however, if you still want to support pre-Dart 2.1, it's totally acceptable to close this PR :)

mogol:
Thanks :+1:

PR #170, opened 2020-10-17
make it a singleton class

mohamedlotfy50:
making (FlutterSecureStorage) class will eliminate creating multiple instance from the same class.
it will make only one instance so is will be easy to use and easy to create a utility class.


GusRodrigues86:
You can wrap your instance of the plugin in your own singleton and let everybody be happy and enjoy a nice const constructor

mogol:
Hey @mohamedlotfy50, 
Thanks for your PR, however I would recommend to use another patterns. Also I do not see limitations to define own singleton for secure storage and access it inside your app. I wouldn't merge it for now.

PR #177, opened 2020-10-28
Fix Locale & KeyPairGeneratorSpec deprecation warnings

0biWanKenobi:
This commit resolves a few deprecation warnings by implementing dedicated methods:

- `config.locale=locale` has new method `setSystemLocaleLegacy  `, otherwise `config.setLocale(locale)`
- `context` is set by new method `setContextConfigurationLegacy` ,otherwise `context = context.createConfigurationContext(config);`
- AlgorithmParameterSpec is obtained with new method `makeAlgorithmParameterSpecLegacy`, otherwise same as before


mogol:
Thanks @0biWanKenobi

PR #181, opened 2020-11-11
Added macOS support.

firatagdas:


macos/.gitignore
```diff
@@ -0,0 +1,31 @@
+.idea/
```
mogol, line 1: Why do we need this file? Wouldn't project level enough?

macos/Classes/FlutterSecureStoragePlugin.m
```diff
@@ -0,0 +1,201 @@
+#import "FlutterSecureStoragePlugin.h"
```
mogol, line 1: It seems pretty close (if not 100% same) to iOS version, can it be reused aka DRY?

talhabalaj: [package_info](https://github.com/flutter/plugins/blob/master/packages/package_info) implements an approach for making code dry between iOS and macOS.

firatagdas: well, i didn’t know that flutter plugins can handle symblinks. It is a good approach though

macos/.gitignore
firatagdas: Flutter adds automatically when you execute flutter create command to add missing platforms

macos/Classes/FlutterSecureStoragePlugin.m
talhabalaj: yeah, I just saw it today.

HeebeLee:
@mogol 
I need this, if you can, please merge as soon as possible. Thank you.

macos/Classes/FlutterSecureStoragePlugin.m
HeebeLee: If the code is consistent with iOS, then this only requires the ln -s command. Of course, we can also share most of the code and do some meticulous work by judging the platform in OC.

I don't understand OC code, so please help to finish this work as soon as possible, thank you.

muhlba91:
any progress on this so far? it's blocking us from porting apps to macos and would appreciate any release or pre-release with macos support.

IonVillarreal:
I also have some applications that I would like to take to the Mac, there is something new about it ?

jhancock4d:
Note that I've broken this out in PR #257 as a separate plugin with the platform interface implemented. Depending on the maintainer's wishes you may want to just work from that PR as it includes web and windows as well.

juliansteenbakker:
I'm closing this because it has been integrated with #255 and #257

PR #184, opened 2020-12-03
Make it possible to have no iOS account name(kSecAttrService)

klyver:
Problem:
Apps migrating from Native to Flutter might be using the SimpleKeychain library or something similar, that uses no account name (kSecAttrService). For such apps it is not possible to share KeyChain with Flutter using this package.

Proposed Solution: 
Make it possible to have no iOS account name(kSecAttrService). Having "flutter_secure_storage_service" as account name is still the default.

### DIFF
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -4,11 +4,13 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:meta/meta.dart';
 
+
 class FlutterSecureStorage {
   const FlutterSecureStorage();
 
   static const MethodChannel _channel =
       const MethodChannel('plugins.it_nomads.com/flutter_secure_storage');
+  static const IOSOptions _defaultIOSOptions = IOSOptions();
 
   /// Encrypts and saves the [key] with the given [value].
   ///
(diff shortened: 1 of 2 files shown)


mogol:
Hey @klyver 
Sorry for late response.
That looks quite useful.

lib/flutter_secure_storage.dart
```diff
@@ -141,12 +141,16 @@ enum IOSAccessibility {
 class IOSOptions extends Options {
   IOSOptions(
       {String groupId,
-      IOSAccessibility accessibility = IOSAccessibility.unlocked})
+      IOSAccessibility accessibility = IOSAccessibility.unlocked,
+      bool useFlutterSecureStorageServiceAsAccountName = true,
```
mogol, line 145: @klyver 
what do you think about  passing AccountName as String with default value `flutter_secure_storage_service`?

klyver: That will probably be better. Then `IOSOptions(accountName: null)` should be allowed and be the way to do it when the user does not want kSecAttrService to be set at all.

klyver:
Hi @mogol - now an accountName String parameter is used instead. 

`accountName` has to be set to `null` to avoid setting `kSecAttrService`.

mogol:
I updated implementation a bit to support null-safety and merged into develop.
Thanks @klyver

PR #187, opened 2020-12-10
null safety

SteveAlexander:
I've made some changes to hopefully make this package null safe

### DIFF
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -140,21 +140,19 @@ enum IOSAccessibility {
 class IOSOptions extends Options {
   IOSOptions(
       {String? groupId,
-      IOSAccessibility? accessibility = IOSAccessibility.unlocked})
+      IOSAccessibility accessibility = IOSAccessibility.unlocked})
       : _groupId = groupId,
         _accessibility = accessibility;
 
   final String? _groupId;
-  final IOSAccessibility? _accessibility;
+  final IOSAccessibility _accessibility;
   @override
   Map<String, String> _toMap() {
     final m = <String, String>{};
     if (_groupId != null) {
       m['groupId'] = _groupId!;
     }
-    if (_accessibility != null) {
-      m['accessibility'] = describeEnum(_accessibility!);
-    }
+    m['accessibility'] = describeEnum(_accessibility);
     return m;
   }
 }
--- a/lib/flutter_secure_storage.dart
+++ b/lib/flutter_secure_storage.dart
@@ -2,7 +2,6 @@ import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
-import 'package:meta/meta.dart';
 
 class FlutterSecureStorage {
   const FlutterSecureStorage();
@@ -20,10 +19,10 @@ class FlutterSecureStorage {
   /// [aOptions] optional Android options
   /// Can throw a [PlatformException].
   Future<void> write(
-          {@required String key,
-          @required String value,
-          IOSOptions iOptions,
-          AndroidOptions aOptions}) =>
+          {required String key,
+          required String? value,
+          IOSOptions? iOptions,
+          AndroidOptions? aOptions}) =>
       value != null
           ? _channel.invokeMethod('write', <String, dynamic>{
               'key': key,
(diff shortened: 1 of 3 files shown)


lib/flutter_secure_storage.dart
```diff
@@ -140,21 +139,21 @@ enum IOSAccessibility {
 
 class IOSOptions extends Options {
   IOSOptions(
-      {String groupId,
-      IOSAccessibility accessibility = IOSAccessibility.unlocked})
+      {String? groupId,
+      IOSAccessibility? accessibility = IOSAccessibility.unlocked})
```
mogol, line 143: @SteveAlexander shouldn't `accessibility` be non-nullable?

lib/flutter_secure_storage.dart
```diff
@@ -140,21 +139,21 @@ enum IOSAccessibility {
...
-  final String _groupId;
-  final IOSAccessibility _accessibility;
+  final String? _groupId;
+  final IOSAccessibility? _accessibility;
   @override
   Map<String, String> _toMap() {
     final m = <String, String>{};
     if (_groupId != null) {
-      m['groupId'] = _groupId;
+      m['groupId'] = _groupId!;
```
mogol, line 153: Need to dig into null-safe dart, but why do we need  `_groupId!` and not `_groupID`?

mogol:
Hi @SteveAlexander 
I added a few comments, but will merge this soon, looks good.

Thanks for your contribution

lib/flutter_secure_storage.dart
hacker1024: I've been using null-safe Dart for a while, and situations like this seem to come up a lot. The inference isn't great.

hacker1024: With sound null safety, I don't think there are any performance hits though.

pubspec.yaml
```diff
@@ -4,8 +4,8 @@ version: 3.3.5
 homepage: https://github.com/mogol/flutter_secure_storage
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
-  flutter: ">=1.20.0 <2.0.0"
+  sdk: ">=2.12.0-0 <3.0.0"
+  # flutter: ">=1.24.0 <2.0.0"
```
koskimas, line 8: This shouldn't be commented out right?

lib/flutter_secure_storage.dart
SteveAlexander: In this case, it's because the code is using a member of the instance, and it could conceivably change or be implemented by a getter that changes the returned value all the time.

So, the analyzer can't be sure that `_groupId` tested in the `if` statement will have the same value by the time we get to the next line.

Solutions — there are discussions in the Dart issue tracker on making the analyzer smarter, or making there be a difference between getters and instance members. But, in the near term, the solution is either to use the `!` or to assign a local variable, instead of using the instance member.

pubspec.yaml
SteveAlexander: that's right! Sorry, can't say why that's there. It looks wrong.

lib/flutter_secure_storage.dart
SteveAlexander: yes!

pubspec.yaml
SteveAlexander: I've changed it to use as a minimum constraint the version of flutter that is on beta channel

lib/flutter_secure_storage.dart
koskimas: That's true, even though not a single person in the known universe would ever implement the getter in a way that would set the value to null, but the analyzer still cannot be sure. Even though it can be 99.9999999999999% sure 😄

SteveAlexander:
I think I've addressed all the feedback on this branch. Thanks to all the reviewers. Anything else that ought to be addressed before merging?

ovestoerholt:
Would love to have a nullsafety version of the library available... Any idea on when this PR may pass?
Thanks, Ove

acoutts:
@mogol any chance to have this merged and a null safety version published soon?

mogol:
Hello folks,
It's merged and published. Give it a try!

Thanks @SteveAlexander

JohnGalt1717:
Is this just 4.0.0 without .nullsafety per the convention or am I missing something?

mogol:
@JohnGalt1717 could you please point to the convention? I was using the latest available guide [here](https://dart.dev/null-safety/migration-guide)

JohnGalt1717:
The null safety docs suggest that package creators use .nullsafety on the end of the versions until the package is stable and then remove it and simply mark it as preview at that point.

Issue #203, opened 2021-02-11
Use EncryptedSharedPreferences instead of SharedPreferences for Android


nilsmagnus:
The current implementation for android uses SharedPreferences, wich will store the key in plaintext. If this is replaced with EncryptedSharedPreferences instead, the keys would also be encrypted. 

About storing data securely on android:
https://developer.android.com/topic/security/data

About EncryptedSharedPreferences:
https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences

nilsmagnus:
I created a gist here to create an instance of EncryptedSharedPreferences:

https://gist.github.com/nilsmagnus/006d9e7ce3b99913df64aa86f8924fe8

matpag:
While I think this is a good proposal, it has 2 problems:
1) Would break compatibility with currently installed keys
2) Security library is still in alpha

mogol:
@nilsmagnus :+1: thanks

what's your opinion on the @matpag 's points?

nilsmagnus:
@mogol / @matpag I agree in both points. It would be a breaking change, so maybe it fits for the next major version and when the the security library is out of alpha?

 However, I need this functionality so I will probably be using a fork of this library as long as the key is not encrypted.

mogol:
One way of implementing is to add an option for initialization, e.g. use EncryptedSharedPreferences or Keystore

nilsmagnus:
@mogol @matpag: I updated the pr with option to use encryptedSharedPreferences. The security-library is still in alpha, but I cannot do anything about that now.

matpag:
You increased minSDK to 23 while seecurity library can operate with minSDK 21, but for API 21-22 library will not use the AndroidKeyStore because it seems a bit unsecure as per the description here: https://developer.android.com/jetpack/androidx/releases/security#security-crypto-1.1.0-alpha01

matpag:
After 2 years Security 1.0.0 is finally stable: https://developer.android.com/jetpack/androidx/releases/security#1.0.0

juliansteenbakker:
This has been implemented in #259

PR #204, opened 2021-02-12
#203 Use EncryptedSharedPreferences instead of raw preferences

nilsmagnus:
This PR enables the usage of EncryptedSharedPreferences instead of the plaintext sharedpreferences. 

Note that this increases the minimum api-level from 18 to 23.


mogol:
It's super breaking changes in current form. 
To merge and publish EncryptedSharedPreferences it should be optional storage engine that new users could choice and existing can migrate to.

If anyone wants to help with these changes, I am happy to discuss how it could be implemented.

juliansteenbakker:
I think it's a good feature however it's not ok to just raise the min SDK because of this. We should implement a function that checks if api-level is lower or equal to/higher than 23  then we should use `EncryptedSharedPreferences`, otherwise fall back to the existing function. Also, it should be an opt-in because people would loose their existing data if we just migrate everyone to `EncryptedSharedPreferences`.@nilsmagnus do you have any time to implement this?

juliansteenbakker:
Closing this in favor of #259

PR #217, opened 2021-03-27
Put the debug console logs behind BuildConfig.DEBUG so that they won't be visible in release builds.

### LINKED ISSUE #214: Read: key exists => Running ensureInitStorageCipher | StorageCipher already initialized
Hello, 
Is it normal to get two debug console log entries every time a secure storage value is read? If I make two consecutive reads, I get 4 log entries; two per read.
```
Read: key exists => Running ensureInitStorageCipher
StorageCipher already initialized
Read: key exists => Running ensureInitStorageCipher
StorageCipher already initialized
```
I noticed this little while ago, but had too many other things I was fixing, and these appear to just be info logs. But I cannot find anything anywhere with these keywords, other than [the actual file in GitHub](https://github.com/mogol/flutter_secure_storage/blob/develop/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java).

Or could it be because I'm using [get_it](https://pub.dev/packages/get_it) and initializing the package above `main`?

Or maybe it's completely normal? If so, is there any way to turn it off?

[main.dart]
```
void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  await setupServiceLocator();
  runApp(MyApp());
}
```
[service_locator.dart]
```
FlutterSecureStorage serviceLocatorSecureStore;
Future<void> setupServiceLocator() async {
  serviceLocatorSecureStore = new FlutterSecureStorage();
```
[storage_service.dart]
```
  Future<String> secureStoreRead(String rkey) async {
    String _tempRead;
    try {
      _tempRead = await serviceLocatorSecureStore.read(key: rkey); // This triggers two log entries.
    } catch (err) {
      _catchErrors('secureStoreRead', err.toString());
    }
    return _tempRead;
  }
```
[Consecutive Reads]
```
final storageKeyID = await _storageService.secureStoreRead('userId');
final storageKeyT = await _storageService.secureStoreRead('userToken');
```
[flutter doctor -v]
```
PS D:\src\projects\kdrecall\hungry> flutter doctor -v
[√] Flutter (Channel stable, 2.0.2, on Microsoft Windows [Version 10.0.19042.867], locale en-US)
    • Flutter version 2.0.2 at D:\src\flutter
    • Framework revision 8962f6dc68 (2 days ago), 2021-03-11 13:22:20 -0800
    • Engine revision 5d8bf811b3
    • Dart version 2.12.1

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2) 
    • Android SDK at C:\Users\micro\AppData\Local\Android\sdk
    • Platform android-29, build-tools 29.0.2
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java       
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe      

[√] Android Studio (version 4.1.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[√] VS Code (version 1.54.2)
    • VS Code at C:\Users\micro\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.20.0

[√] VS Code, 32-bit edition
    • VS Code at C:\Program Files (x86)\Microsoft VS Code
    • Flutter extension version 3.20.0

[√] Connected device (3 available)
    • Pixel XL (mobile) • HT74V0204689 • android-arm64  • Android 10 (API 29)       
    • Chrome (web)      • chrome       • web-javascript • Google Chrome 89.0.4389.82
    • Edge (web)        • edge         • web-javascript • Microsoft Edge 89.0.774.45

• No issues found!
```
Thanks for any insights... I would have posted on SO, but unsure if this is normal or not, and if not, figured it'd be better to post at the source. :) However, I can if need be. -Keith

PieterAelse:
Fixes #214 

I did not went for any customization Logging options on Dart side because that would be more changes and I believe this is enough to fix the logs that I've introduced earlier.


android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -70,14 +71,14 @@ public void initInstance(BinaryMessenger messenger, Context context) {
     private void ensureInitStorageCipher() {
         if (storageCipher == null) {
             try {
-                Log.d(TAG, "Initializing StorageCipher");
+                if (BuildConfig.DEBUG) Log.d(TAG, "Initializing StorageCipher");
```
mogol, line 74: Could we simply remove that?

PieterAelse: Sure, only this print or also the others?

drmirk: I think you should remove all prints altogether. Mundane tasks should not print anything. But instead there could be a switch or something(I don't know how hard it is), which user can use to enable/disable printing/logging.

KDCinfo: I'm still new to Flutter and Dart (and mobile and OOP). Are there different layers of logging; info, warn, error?

mogol: @PieterAelse I would suggest to remove all prints. Agree with @drmirk. 

Ideally, the plugin should receive a parameter (probably, during initialization) with required Log level, which is off by default. Should not be hard, except to make sure it's consistent across all platforms.

PieterAelse: @mogol PR is updated, debug logs have been removed. Did not touch error logs (on crashes).

mogol:
Thanks @PieterAelse ,  merged!

chitgoks:
this will be in 4.1.1?

EdsonBueno:
I need this too

PR #244, opened 2021-06-23
docs: added info on IOSOptions

loonix:
added missing information that is hard to find unless you check the actual plugin file


juliansteenbakker:
Thanks for your contribution!

PR #246, opened 2021-06-27
Replace jcenter with mavenCentral

0biWanKenobi:
Since jcenter is officially EOL, and using it results in a compile time warning, it is [recommended](https://developer.android.com/studio/build/jcenter-migration) to switch to mavenCentral.


juliansteenbakker:
Thanks for your contribution!

PR #255, opened 2021-07-23
Added macOS support based on #181

juliansteenbakker:
This MR is based on #181, but updated with the latest code from the plugin, an example project, and a fix for macOS 10.15 and above.

After some testing i found out that on macOS 10.15 and above, we need to add the following to the write function:
```objc
    if (@available(macOS 10.15, *)) {
        search[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
    }
```

After adding this and the correct entitlements to the project, namely Keychain Sharing, everything works as expected. 

@mogol With this MR, we can close the other two existing macOS MR's.


juliansteenbakker:
Implemented in #257

PR #257, opened 2021-07-23
Update to use platform interface and implement web, windows, macOS and split Linux into it's own project per best practices guidance from Google

jhancock4d:
Per: #252 here's the PR to merge this and implement web with full platform interface support. I also incidently fixed a bug in the iOS example project todo with the plugin name space. And I implemented macOs and windows options ready for those PRs to be merged (which could be their own projects as could the linux one now if you so desire to make them function that way as suggested by the flutter team for plugin development)

I have tested that I didn't break anything on Android, and tested the web functionality which appears to be done correctly.  It is my understanding from the referenced white papers that this results in the private key still be stored in the browser. I have tested this between browsers with the information that is stored in localstorage and I was not able to reverse the encryption on a second browser.

PR #262, opened 2021-07-29
imp: changed to endorsed plugin

juliansteenbakker:
After some testing and comparing with [the shared_preferences package] (https://github.com/flutter/plugins/tree/master/packages/shared_preferences) that is also a federated plugin, i discovered that the current implementation (#257) is a so called non endorsed implementation. This means that you have to manually add each type of package (i.e. flutter_secure_storage_macos) to your pubspec.yaml. I don't think this is the way to go, so i have converted the package to an endorsed plugin. This means that you only have to import `flutter_secure_storage` like you normally would, and it would automatically include all the platforms needed.
Also, i have moved the example to the `flutter_secure_storage` and removed unused dart files from various platform implementations.

I am currently still testing all platforms so this is still a draft.

@jhancock4d What do you think about this? Was there any reason behind making a non federated project?


jhancock4d:
Only ignorance on my part. I wasn't aware that there is a difference. What specifically did you change to make it federated?

juliansteenbakker:
The key here is to add the following to the pubspec.yaml of all the platform packages:
```dart
plugin:
    implements: flutter_secure_storage
```

This means that when i upload for examle flutter_secure_storage_macos, it know's that it shouldn't visibly show it on pub.dev, but rather that this package is a part of the flutter_secure_storage interface plugin.

jhancock4d:
Does that implement the platform interface or the plugin with android and iOS?

jhancock4d:
Might want to consider releasing a development build to pub.dev with this code as a beta if you've tested across all platforms so that we can get feedback. Only thing that is really a blocker other than bugs in my mind from this going to the production release as a new major version is the Windows implementation needing readAll and deleteAll implemented.

jhancock4d:
@juliansteenbakker Note that the web documentation in the readme references adding it directly. It needs to be fixed with this merge so that it doesn't tell you to do that. Likely the same on the readme for the actual web project and may be an issue with other platforms in the docs.

juliansteenbakker:
Thanks for mentioning it, i have changed the readme. I already edited all the readme of the separated projects like so: 

> Usage
> This package is endorsed, which means you can simply use flutter_secure_storage normally. This package will be automatically included in your app when you do.

jhancock4d:
Ok, the main readme that is displayed on github when you go to the root of this project under web is wrong then. I can do a pull to fix if you want or you can just update it with something else if that works better.

juliansteenbakker:
Can you give me a link to the readme? I can't seem to find a wrong readme file.

jhancock4d:
https://github.com/mogol/flutter_secure_storage/tree/develop/flutter_secure_storage

Basically there has to be the same readme in both the root and the actual project. The first so that people that land on git hub get the right one, and the second so that if you land on pub.dev you get the right one. If you have a better solution, I'd love to see it. Kinda annoying have to keep 2 up-to-date.

juliansteenbakker:
Aah i see, hmm well i don't think there is a better solution. I will update the readme right now.

PR #271, opened 2021-08-02
Fixes Secure Storage Web Bug on containsKey and implements a containsKey option in the example app that could be integration tested.
Updates files that are required for linux etc. that were not in the repo for Linux. (confirmed not to be in .gitignore)

jhancock4d:


jhancock4d:
This fixes the containsKey issue and puts in a containsKey option in the example app on the ... of each key which is obviously redundant because it exists, but this allows an integration test to be written against that option.

Note that android, ios and several others still don't implement containsKey properly.

flutter_secure_storage_platform_interface/pubspec.lock
```diff
@@ -70,7 +70,7 @@ packages:
       name: charcode
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "1.3.1"
+    version: "1.2.0"
```
juliansteenbakker, line 73: It seems like it's downgrading packages in the interface package even while you didn't really change anything in the interface package. Have you run `flutter pub upgrade`?

JohnGalt1717:
Flutter pub upgrade only upgrades mokito when run.  Which doesn't change the version of charcode that you're referencing above. It' still at 1.2.0.  Only thing I can't think of other than below is you're using a flutter beta version instead of stable?

This might be caused by the publish_to because you have a dependancy to a library version that doesn't yet exist on pub.dev.  (Google has issues with this, becuase if you update a dependancy to relative, you can't publish, but you can't test without it being published, so you have to change publish_to to none and then relative path it, and then change it back after you're done testing before committing which causes issues)

flutter_secure_storage_platform_interface/pubspec.lock
jhancock4d: Flutter pub upgrade only upgrades mokito when run. Which doesn't change the version of charcode that you're referencing above. It' still at 1.2.0. Only thing I can't think of other than below is you're using a flutter beta version instead of stable?

This might be caused by the publish_to because you have a dependancy to a library version that doesn't yet exist on pub.dev. (Google has issues with this, becuase if you update a dependancy to relative, you can't publish, but you can't test without it being published, so you have to change publish_to to none and then relative path it, and then change it back after you're done testing before committing which causes issues)

juliansteenbakker: I'm currently running stable, but it's no big deal anyway.

PR #274, opened 2021-08-04
Windows implementations of `containsKey`, `readAll` and `deleteAll`

kennethnym:
This PR includes Windows implementations of `containsKey`, `readAll` and `deleteAll` operations.

This PR also addresses #275.

In the example folder, I have to specify in `pubspec.yaml` to override `flutter_secure_storage_windows` to use the package locally instead of pulling from pub.dev, because local changes aren't be reflected in the example app otherwise. If this should be the desired behavior, all `flutter_secure_storage_*` dependencies should point to local folders as well.

~~For `containsKey` implementation, since `read` returns `null` when the given key doesn't exist in storage, implementing it in Dart should be sufficient:~~

Edit: @jhancock4d made a good point that null checks aren't sufficient to determine whether a value exists in storage. This PR will also include `containsKey` implementation.

### DIFF
--- a/flutter_secure_storage/example/pubspec.yaml
+++ b/flutter_secure_storage/example/pubspec.yaml
@@ -24,5 +24,9 @@ dev_dependencies:
     sdk: flutter
   test: any
 
+dependency_overrides:
+  flutter_secure_storage_windows:
+    path: ../../flutter_secure_storage_windows
+
 flutter:
   uses-material-design: true
--- a/flutter_secure_storage/example/pubspec.yaml
+++ b/flutter_secure_storage/example/pubspec.yaml
@@ -24,9 +24,5 @@ dev_dependencies:
     sdk: flutter
   test: any
 
-dependency_overrides:
-  flutter_secure_storage_windows:
-    path: ../../flutter_secure_storage_windows
-
 flutter:
   uses-material-design: true
--- a/flutter_secure_storage/pubspec.yaml
+++ b/flutter_secure_storage/pubspec.yaml
@@ -35,7 +35,7 @@ dependencies:
   flutter_secure_storage_linux: ^1.0.0
   flutter_secure_storage_macos: ^1.0.0
   flutter_secure_storage_platform_interface: ^1.0.0
-  flutter_secure_storage_web: ^1.0.0
+  flutter_secure_storage_web: ^1.0.1
   flutter_secure_storage_windows: ^1.0.0
 
 dev_dependencies:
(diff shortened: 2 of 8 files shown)


juliansteenbakker:
> In the example folder, I have to specify in pubspec.yaml to override flutter_secure_storage_windows to use the package locally instead of pulling from pub.dev, because local changes aren't be reflected in the example app otherwise. If this should be the desired behavior, all flutter_secure_storage_* dependencies should point to local folders as well.

Well, this is not how federated plugins work when publishing. Every platform integration has a separate pub.dev entry and as such, it can't point to a local dependency. When developing on a federated plugin, it is indeed necessary to use the separate packages locally. However when merging and submitting this for pub.dev (and for this repository) the correct way is to bump the version of the package you have edited and add a changelog entry with the changes you made. When this is merged, i can first upload the changed platform integration, and after that update the example app with the new version number we just uploaded.

So in short: You should bump windows version from 1.0.0 to 1.1.0 (new feature, otherwise it would be 1.0.1) and add to the changelog what you have added.

flutter_secure_storage/pubspec.yaml
```diff
@@ -35,7 +35,7 @@ dependencies:
   flutter_secure_storage_linux: ^1.0.0
   flutter_secure_storage_macos: ^1.0.0
   flutter_secure_storage_platform_interface: ^1.0.0
-  flutter_secure_storage_web: ^1.0.1
+  flutter_secure_storage_web: ^1.0.0
```
juliansteenbakker, line 38: Why did you downgrade the web version?

flutter_secure_storage/example/pubspec.yaml
```diff
@@ -24,5 +24,9 @@ dev_dependencies:
     sdk: flutter
   test: any
 
+dependency_overrides:
```
juliansteenbakker, line 27: As commented, please only use override when developing locally but don't commit this to this repository.

juliansteenbakker (changes requested):

flutter_secure_storage/pubspec.yaml
kennethnym: Sorry, I didn't explain this in the original post. I was having an issue with version solving, but today I reverted the change and everything works fine. My apologies.

flutter_secure_storage/example/pubspec.yaml
kennethnym: Will do. Thank you for the detailed explanation,

juliansteenbakker (approved):

PR #281, opened 2021-08-09
Add option defaults at the overall class

jhancock4d:
This sets defaults for the overall class that eliminates bugs and risk of repeating the same options per call.  This also facilitates dependency injection sharing options.


juliansteenbakker:
Thanks for your contribution!

jhancock4d:
@juliansteenbakker You might consider removing the options from the actual calls in that class as a breaking change for V5, because I think it is poor design to have them there and honestly can't think of a case where you wouldn't set this globally nor want to override the values for the options....

Issue #294, opened 2021-08-22
BIG ISSUE ( WINDOWS ) : It shares the data between two apps with the same key


topperspal:
### STEPS TO RECREATE

1. Create two different apps with `flutter_secure_storage 5.0.0-beta.4.`
2. Open First app and save data using FlutterSecureStorage().write(key: "mykey", value: "my_highly_secured_key")
3. Open Second app and try to read the data using `FlutterSecureStorage().read(key: "mykey")`

You will be able to read the my_highly_secure_key from another app.

PLEASE FIX IT SOON

juliansteenbakker:
This is fixed in #325 and will be released soon.

topperspal:
If two apps are named same, they can read creds of each other :(. Please reopen this issue.

I have edited `STEPS TO RECREATE` above, please go through them :)

PR #325, opened 2021-11-09
[windows] Make credential key unique to the app

Jon-Salmon:
Resolves issue #294 

I also added dependency overrides to the example app to simplify local plugin testing. Can be removed if not desired.

### DIFF
--- a/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp
+++ b/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp
@@ -65,8 +65,8 @@ namespace
     bool ContainsKey(const std::string &key);
   };
 
-  const std::string ELEMENT_PREFERENCES_KEY_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIHNlY3VyZSBzdG9yYWdlCg_";
-  const int ELEMENT_PREFERENCES_KEY_PREFIX_LENGTH = 55;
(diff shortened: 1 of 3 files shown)


flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp
```diff
@@ -65,8 +65,8 @@ namespace
     bool ContainsKey(const std::string &key);
   };
 
-  const std::string ELEMENT_PREFERENCES_KEY_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIHNlY3VyZSBzdG9yYWdlCg_";
-  const int ELEMENT_PREFERENCES_KEY_PREFIX_LENGTH = 55;
+  const std::string ELEMENT_PREFERENCES_KEY_PREFIX = SECURE_STORAGE_KEY_PREFIX;
+  const int ELEMENT_PREFERENCES_KEY_PREFIX_LENGTH = (sizeof SECURE_STORAGE_KEY_PREFIX) - 1;
```
juliansteenbakker, line 69: Out of curiosity, why the - 1 at the length?

Jon-Salmon: In c++, sizeof returns the string length rather than the char count and this includes the null string terminator.

juliansteenbakker: Thanks for the explanation!

juliansteenbakker:
Thanks for your contribution!

Issue #307, opened 2021-09-08
Windows issue on writing or reading values


JsonLinesCode:
I'm encountering issues each time I interact with Windows flutter_secure_storage.
It seems to be new, maybe related to the last release ?

The returned issue is : 
`PlatformException(Exception encountered: UNKNOWN_ERROR, delete, null, null)
`
I've reproduced the issue on several Windows devices.

florian-lefebvre:
On windows, it seems that using `write` with a null value is responsible of this error, as well as using `delete` when key doesn't exist. I managed to get it work with a wrapper class:
```dart
import 'package:flutter_secure_storage/flutter_secure_storage.dart';

/// Stands for Key Value Store. Uses [FlutterSecureStorage] under the hood.
class KVS {
  static const _storage = FlutterSecureStorage();

  /// Stands for Key Value Store. Uses [FlutterSecureStorage] under the hood.
  const KVS._();

  static Future<String?> read({required String key}) async {
    return await _storage.read(key: key);
  }

  static Future<void> write({required String key, required String value}) async {
    return await _storage.write(key: key, value: value);
  }

  static Future<void> delete({required String key}) async {
    if (await containsKey(key: key)) {
      return await _storage.delete(key: key);
    }
  }

  static Future<bool> containsKey({required String key}) async {
    return await _storage.containsKey(key: key);
  }

  static Future<Map<String, String>> readAll() async {
    return await _storage.readAll();
  }

  static Future<void> deleteAll() async {
    return await _storage.deleteAll();
  }
}
```

juliansteenbakker:
This is fixed in v5.0.0

PR #326, opened 2021-11-09
[windows] Silently ignore errors when deleting keys that don't exist

Jon-Salmon:
Resolves issue #307


juliansteenbakker:
Thanks for your contribution!

PR #367, opened 2022-02-20
Update the docs of FlutterSecureStorage.delete().

YRFT:
In some cases, the keys not existing will be given to the `FlutterSecureStorage.delete()` function.

In my opinion, it's better to make this scenario clear.


YRFT:
OK. I will do it in a few days.

Issue #374, opened 2022-03-01
Android - Exposing the plugin native API to the native part of the app


pdegand:
Hi,

I would like to call the native part of the plugin directly from the native part of my app (from a widger provider to be precise).

As I don't wan't to "copy" the implementation of the plugin, I'm wondering if it would be possible to have all the "core" logic of the plugin in a Flutter-free "core"  and have all the Flutter glue calling that core. This way, my native app could call only the native "core" and not worrying about "faking" the Flutter classes required to call `initInstance` and `onDetachFromEngine`.

This "core" module could be single threaded and it's either on the Flutter glue or on the other callers, such as the native part of any app, to worry about thread switching for performance.

I can work on a proposal implementation of such a solution. Is this something that you would be willing to integrate into this plugin ? 

Thank you.

pdegand:
I did a first implementation on #375. Let me know what youu think of this.

PR #375, opened 2022-03-02
#374 - Expose the Android API used by the Flutter plugin

pdegand:
This PR propose an implementation for #374 .

The goal is to have a public standalone object that does all the work about shared preferences and encryption and to simplyy use this object in the Flutter plugin. This enable native code from the host app to call the same object used by the plugin to read/write the secured storage, just as if it was called from Flutter.

I this first implementation, I simply moved (copy/past ...) most of the Android-specific code into a new class to keep in the plugin only the stuff that is specific to the bridge with Flutter.


juliansteenbakker:
Thanks for your contribution! I have just one question: could you rename the class and filename to `FlutterSecureStorage` instead of `SecureStorageAndroid`?

pdegand:
Thank you for the feedback, I update  the PR with the rename.

PR #395, opened 2022-05-15
WIP: Migrated to EncryptedSharedPreferences

mogol:
Hey folks,
I want to test implementation for Android with only EncryptedSharedPreferences.
I would expect it to be more stable 🤷‍♂️ as the logic is much simpler comparing with legacy.

Drawbacks: 
* Android Min SDK 18 => 23.
* Not Backward Compatible
* Testing in Progress

I am testing it within my projects. I would appreciate it if you could try itry by referencing this branch directly in your pubspec.

```
+  flutter_secure_storage:
+    git:
+      url: https://github.com/mogol/flutter_secure_storage.git
+      ref: migrate-encrypted-preferences
```


krzysztof-swi:
I was thinking about migrating to EncryptedSharedPreferences in my projects, however I found a few threads that there are random problems with it on different devices. Current implementation works fine for me, so I would love to have this alternative at least while the issue with EncryptedSharedPreferences isn't fixed by Google. Please see link below:
[https://issuetracker.google.com/issues/176215143](url)

mogol:
@juliansteenbakker thanks for sharing, it makes sense to wait until it's ready for production 👍 quite sad

mogol:
@juliansteenbakker @krzysztof-swi   I've gone through the links you shared once more.

It looks like the issue is with Keystore itself, not with EncryptedSharedPreferences. So we might still consider fully migrating to EncryptedSharedPreferences and applying workarounds from the links with auto cleanup if things go wrong.

* Unify and simplify the package 
* Apply and reuse workarounds (from the shared links) for everyone that are expected to increase stability (not fully, but a lot)
* Anyway the problem exists in the Keystore and the base solution inherits it as well 😢 
 
Otherwise given the high number of issues and complaints regarding Android 🤦, another solution should be considered. But it might be a lot of work.

Wdyt?

juliansteenbakker:
@mogol  I think the only way to keep everyone happy is to offer EncryptedSharedPreferences (as default?), but with the option to 'old' shared preference with manual encryption. The keystore problem only exists with the EncryptedSharedPreferences library so this will always generate some complaints with implementation. If we would take away the option to use the 'old' shared preference, people who rely on stability will stop using this package (I think).

AlexDochioiu:
@juliansteenbakker I am one of the people that started encountering stability issues caused by `EncryptedSharedPreferences` despite running a pre-release beta with only 50 users. That pretty much makes it clear to me that I have to revert to manual encryption. Removing this option indeed will force me to find another alternative that allows it or just never update the plugin for as long as possible.

Issue #256, opened 2021-07-23
Migrate existing objc code to Swift


juliansteenbakker:
Right now all the iOS and macOS code is written in objc. However, i think it would be better if we migrate this code to swift since it has a cleaner syntax, and thus will be easier to read and maintain. Also, apple actively promotes using Swift over objc and Flutter also uses it as a default language when creating new plugins or projects.

juliansteenbakker:
Fixes in v6.1.0-beta.1

PR #398, opened 2022-05-17
fix issue of memory leak

devinzhao123:
Hello author, first of all, thank you very much for providing this useful plugin. I'm having some memory leak issue while using this plugin. 
image
I use xcode's instrument and found that it is the problem of 'read()' and 'readAll()'. It looks like the cause of the memory leak is that the memory of the CFDataRef is not freed.
image
After I released the memory of CFDataRef and tested it, there was no memory leak issue.
image
Thank you for reviewing my PR and to find better solutions. Hopefully a new plugin version can be released to solve my memory leak issue.


juliansteenbakker:
Thank you very much for you contribution. As stated in #256 my wish is to migrate all of this obj c code but i don't have the knowledge for now.

Issue #356, opened 2022-01-18
include file: 'atlstr.h' is missing


eae1212:
Error during Building Windows application:
fatal error C1083: Cannot open include file: 'atlstr.h': No such file or directory [XXX\_TESTS\flutter_secure_storage_example\build\windows\plugins\flutter_secure_storage_windows\flutter_secure_storage_windows_plugin.vcxproj]

Flutter 2.5.3
flutter_secure_storage: ^5.0.2

eae1212:
Error occurs when building example on Windows 10.
Building on Windows 7 is Ok

alexkmbk:
> @alexkmbk and @eae1212 can you guys attach the link for these resources to be downloaded

https://visualstudio.microsoft.com/downloads/?q=build+tools

PR #411, opened 2022-06-28
Add a note on the C++ ATL requirement for windows

farcaller:
Based on #356.


RedstoneMedia:
Did this get removed ? The note is not present anymore

PR #424, opened 2022-08-04
Imp/linter

juliansteenbakker:


tenhobi:
Hey, can I ask you why did you remove `public` for `String read(String key)` function? In https://github.com/mogol/flutter_secure_storage/pull/375 (related to https://github.com/mogol/flutter_secure_storage/issues/374) it was made so we can share the access to Flutter Secure Storage from native code?

tenhobi:
Right now, I could potentially read the data using `FlutterSecureStorage().readAll()`, which is `public`, but the fun part is that `options` are only set using the plugin, and since I do not use plugin from native part, it is not set and it throws.

Issue #462, opened 2022-10-12
Wrong type for groupId parameter on iOS native implementation


aloys-c:
Hi !

Planned to use the plugin with iOS App Groups to access the data from an extension. Yet, I just noticed the implementation for the groupId parameter on the native side is of type Int? when it should be a String passed to "ksecattraccessgroup", is it indeed a mistake or am I missing something ?

Best wishes

juliansteenbakker:
This is fixed with version 6.1.0

PR #465, opened 2022-10-16
groupId type updated

aloys-c:
Related to Issue [#462](https://github.com/mogol/flutter_secure_storage/issues/462)

The groupId variable should be of type String? to support keychain sharing and app groups as the identifiers are in the form of "com.company.appid.keychain"... The current type of Int doesn't produce a compilation error as the main group is selected in case an invalid argument is provided, yet without this correction it's not possible to access the data from extensions and app groups.

Tested and works fine with the Notification Service Extension.


busterc:
+1

juliansteenbakker:

PR #488, opened 2022-12-12
add GCM instead of CBC

beshoy97:
fix security issue because CBC encryption not secure:
https://github.com/MobSF/owasp-mstg/blob/master/Document/0x04g-Testing-Cryptography.md#identifying-insecure-andor-deprecated-cryptographic-algorithms-mstg-crypto-4


juliansteenbakker:
You can already use GCM using the AndroidOptions parameter. See https://github.com/mogol/flutter_secure_storage/blob/develop/flutter_secure_storage/lib/options/android_options.dart

beshoy97:
I know, but when using a security tool gives an alert of using CBC encryption in getCipher method :
https://github.com/beshoy97/flutter_secure_storage/commit/1fab5256f51fa8d38227408f61cb2af393d83295
CBC encryption vulnerabilities:
https://learn.microsoft.com/en-us/dotnet/standard/security/vulnerabilities-cbc-mode

PR #495, opened 2022-12-27
Linux: fix memory management issues

jpnurmi:


juliansteenbakker:
Thank you for your contribtion. I however can't test this implementation. Have you been able to test this with multiple devices?

Issue #450, opened 2022-09-09
[Windows] Unable to store data more than 2559 characters. Errors with "UNKNOWN ERROR"


ThomasAunvik:
```dart
final anotherData = Iterable.generate(2559, (_) => "a").join();
await storage.write(key: "data", value: anotherData + "a");
```

Remove the extra "a" to successfully store.

```
PlatformException(Exception encountered: UNKNOWN_ERROR, write, null, null)
StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:653:7)
MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:296:18)
<asynchronous suspension>
```

Could it be flutter that is the cause?

flutter_secure_storage: 5.1.2
<details>
  <summary>Flutter Doctor -V</summary>

```
Flutter Info:
[✓] Flutter (Channel stable, 3.3.0, on Microsoft Windows [Version 10.0.22000.856], locale nb-NO)
    • Flutter version 3.3.0 on channel stable at C:\Users\thaun\Documents\FlutterSDK
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ffccd96b62 (11 days ago), 2022-08-29 17:28:57 -0700
    • Engine revision 5e9e0e0aa8
    • Dart version 2.18.0
    • DevTools version 2.15.0

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at C:\Users\thaun\AppData\Local\Android\sdk
    • Platform android-31, build-tools 31.0.0
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[✓] Visual Studio - develop for Windows (Visual Studio Enterprise 2022 17.3.3)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Enterprise
    • Visual Studio Enterprise 2022 version 17.3.32825.248
    • Windows 10 SDK version 10.0.19041.0

[✓] Android Studio (version 2020.3)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
[...]

Issue #361, opened 2022-02-02
Exception has occurred. PlatformException (PlatformException(Exception encountered: UNKNOWN_ERROR, write, null, null))


insinfo:
i'm getting this error when trying to store a large object on Windows

```dart
class Usuario implements ModelBase {
  @override
  String getClassName() {
    return 'Usuario';
  }

  Usuario({
    this.id,
    required this.username,
    required this.password,
  });

  Usuario.fromMap(Map<String, dynamic> map) {
    id = map.containsKey('id') ? map['id'] : id;
    username = map.containsKey('username') ? map['username'] : username;
    password = map.containsKey('password') ? map['password'] : password;
    ativo = map.containsKey('ativo') ? map['ativo'] : ativo;
    idPerfil = map.containsKey('idPerfil') ? map['idPerfil'] : idPerfil;
    idPessoa = map.containsKey('idPessoa') ? map['idPessoa'] : idPessoa;
    dataCadastroAsString = map.containsKey('dataCadastro')
        ? map['dataCadastro']
        : dataCadastroAsString;

    if (map.containsKey('perfil')) {
      if (map['perfil'] is Map<String, dynamic>) {
        perfil = Perfil.fromMap(map['perfil'] as Map<String, dynamic>);
      }
    }
    if (map.containsKey('pessoa')) {
      if (map['pessoa'] is Map<String, dynamic>) {
        pessoa = Pessoa.fromMap(map['pessoa'] as Map<String, dynamic>);
      }
    }
    if (map.containsKey('accessToken')) {
      accessToken = map['accessToken'];
    }
    if (map.containsKey('expiresIn')) {
      expiresIn = DateTime.parse(map['expiresIn']);
    }

    if (map.containsKey('usuarioJubarte')) {
      usuarioJubarte = UsuarioJubarte.fromMap(map['usuarioJubarte']);
    }
  }

  int? id;
  late String username;
  String? password;
  bool ativo = true;
  int? idPerfil;
  int? idPessoa;
  DateTime? dataCadastro;
  String? accessToken;
  DateTime? expiresIn;
  UsuarioJubarte? usuarioJubarte;

  String get dataCadastroAsString {
    final dt = dataCadastro != null
        ? dataCadastro!.toIso8601String().substring(0, 10)
        : '';
    return dt;
  }

  set dataCadastroAsString(dynamic value) {
    if (value is DateTime) {
      dataCadastro = value;
    } else if (value is String) {
[...]

Issue #345, opened 2021-12-03
UNKNOWN_ERROR on windows for large secrets (>2561 bytes)


volund:
The following error occurs on windows whenever I try to save a secret that's larger than 2561 bytes:

```
    PlatformException(Exception encountered: UNKNOWN_ERROR, write, null, null)
    
    #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:607:7)
    #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:156:18)
    <asynchronous suspension>
```

This error does not appear with a secret that is 2560 bytes or less. This is happening with 5.0.2 on Windows 10 Enterprise. 

For reference here's the secret I reproduce the error with:

```
[...]

PR #505, opened 2023-01-15
[Windows] Write encrypted data to files instead of the windows credential system

Mino5531:
The data to store is encrypted with a key that is stored in the windows credential system and then written to a file in the app specific `%appdata%` folder. 
This fixes: #450, fixes: #361, fixes: #345

This should also be backwards compatible with old data stored in the windows credential system.

### DIFF
--- a/flutter_secure_storage_windows/windows/CMakeLists.txt
+++ b/flutter_secure_storage_windows/windows/CMakeLists.txt
@@ -22,5 +22,8 @@ set(flutter_secure_storage_bundled_libraries
   ""
   PARENT_SCOPE
 )
-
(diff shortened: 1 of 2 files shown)


juliansteenbakker:
Thank you for your work. I need to dust of my windows laptop in order to check it, but afterwards ill let you know.

juliansteenbakker:
I hope that someone else has a windows pc to test this on, because i have some problems firing up mine.

cwangfr:
i have tested. it works.

Eradparvar:
Thank you @juliansteenbakker  for doing a great job maintaining the package.  What are your thoughts about merging this into the default branch?

PR #508, opened 2023-01-18
use preferences fallback when encryption fails

### LINKED ISSUE #380: Attempt to invoke interface method 'boolean android.content.SharedPreferences.contains(java.lang.String)'
`PlatformException(Exception encountered, read, java.lang.NullPointerException: Attempt to invoke interface method 'boolean android.content.SharedPreferences.contains(java.lang.String)' on a null object reference
	at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin$b.run(SourceFile:24)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:246)
	at android.os.HandlerThread.run(HandlerThread.java:67)
, null)`

This issue was supposed to be fixed with #332 , but is still hapenning in 5.0.2

Only Android devices are affected by this issue.

jonasbark:
Should fix #380

### DIFF
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
@@ -151,11 +151,11 @@ public class FlutterSecureStorage {
         if (getUseEncryptedSharedPreferences()) {
             try {
                 preferences = initializeEncryptedSharedPreferencesManager(applicationContext);
+                checkAndMigrateToEncrypted(nonEncryptedPreferences, preferences);
             } catch (Exception e) {
                 Log.e(TAG, "EncryptedSharedPreferences initialization failed", e);
+                preferences = nonEncryptedPreferences;
             }
-
-            checkAndMigrateToEncrypted(nonEncryptedPreferences, preferences);
         } else {
             preferences = nonEncryptedPreferences;
         }


benjaminbisinger:
I'm not sure if this is the best fix for https://github.com/mogol/flutter_secure_storage/issues/380.

In our case, the encrypted preferences can not be created when the app restores an app data backup. This is the default behavior for any app by the way, and must be disabled by using [backup rules](https://developer.android.com/guide/topics/data/autobackup). I think that this case never works, because the backed up preferences can not be decrypted anymore. 

A better solution in this case could be to delete the file and create new, encrypted preferences. Falling back to non encrypted would weaken the security.

If the source of the problem is always a backup, adding the backup exclusion in the manifest of this lib could also be a solution.

jonasbark:
I generally agree that an unencrypted fallback is not desired, but this is what the plugin author originally intended to do (at least that's my guess).

This PR will fix the NullPointerExceptions that are mentioned in the issues OP.

TheCarpetMerchant:
This PR goes completely against what the package "advertises" itself as. If it because unencrypted because of a random bug, it should not be used at all.
Making something unencrypted because something fails with the encryption is the worse possible "fix". It's not fixing anything, it's putting the end user in danger and is completely irresponsible.

jonasbark:
Don't target your hate at me - I'm not responsible for the bug. If I'm misinterpreting what the Author initially intended to achieve then he can simply close this PR. I don't care.

Looking forward to your PRs, then. I'm sure everyone will be very happy with a better solution from you.

jonasbark:
And for what it's worth: the checkAndMigrateToEncrypted is still called after the first successful call to initializeEncryptedSharedPreferencesManager - which will encrypt your unencrypted preferences and thus will do what it's advertised to do. Albeit with a delay. 

I would've solved it very differently, but then again I don't have all the background information from the author.

juliansteenbakker:
First of all, this package always encrypts data. `nonEncryptedPreferences` means that we are just using the default sharedPreferences instead of the android crypto package. This does NOT mean the data is not encrypted, because we encrypt this ourself before storing it. The difference with the android crypto package is that the shared preferences itself is encrypted, so the data we store should not have to be. Otherwise it will be a double encryption.

As for the fix, thanks!

atomancok:
I think there is still a problem with this fix.

Currently `getUseEncryptedSharedPreferences()` ist used to determine if
- if `encrypetedSharePreferences` are used or unencrypted `sharedPreferences` are used
- if the plugin encrypts/decrypts the data itself.

In the case of the implemented fallback, an unencrypted sharedPreferences storage will be used, but because `getUseEncryptedSharedPreferences()` is still true, nothing is encrypted additionally. (See lines 97 and 100 for example).

Please check if my assumptions are correct and consider reverting this fix, if sensitive data is stored unencrypted because of it. I can also prepare a PR that follows the intention of this solution but fixes the describes issue.

juliansteenbakker:
Hi @atomancok, thanks for your insights. I have refactored some code regarding this issue and will publish this later today.

justincbeck:
When will this hit https://pub.dev?

PR #536, opened 2023-03-07
Perform Keychain operations asynchronously on a non-main thread.

adamgic:
Change to perform Keychain operations asynchronously on a non-main thread.

Apple suggest in documentation, that Keychain operations should be called from non-main thread to avoid undesired/unpredictable behaviour on the UI side:

quote
**Performance considerations**
SecItemCopyMatching blocks the calling thread, so it can cause your app’s UI to hang if called from the main thread. Instead, call SecItemCopyMatching from a background dispatch queue or async function.
unquote
https://developer.apple.com/documentation/security/1398306-secitemcopymatching


juliansteenbakker:
Thanks for your change, LGTM!

devmil:
@juliansteenbakker  Is there already a plan for when this will be released?

PR #537, opened 2023-03-15
Enhance security from AES-128 to AES-256

tianli91:
There are many security guidelines and standards that recommend or require the use of strong encryption algorithms such as AES-256. For example, the National Institute of Standards and Technology (NIST) recommends using AES with a key size of at least 128 bits for protecting sensitive information. Similarly, other organizations and standards such as PCI DSS (Payment Card Industry Data Security Standard) also require the use of strong encryption algorithms like AES-256 for protecting sensitive data.


juliansteenbakker:
Won't this break the current implementation?

PR #538, opened 2023-03-15
Add macOS App Groups Documentation

HankG:
I was having problems getting this working with a macOS app that was using App Groups. These new added instructions are how I was able to finally get this working. This may also be an issue with iOS but I haven't tested under those conditions.


juliansteenbakker:

PR #540, opened 2023-03-20
FFI based Windows implementation to solve key related issues

### LINKED ISSUE #539: Some kinds of keys cause app crash in Windows
I found that some kinds of keys cause various troubles in Windows. Specifically,

* Keys with special charactors causes following problems:
  * Keys cannot contain chars which cannot be used in file name, such as `:`. A file without `.secure` extension is created, but it never be read.
  * Keys containing `:/` causes app crash, rather than throwing `PlatformException`. 
    * For example, [oauth2_client](https://pub.dev/packages/oauth2_client) caches access token using URL, which contains ':/'.
  *  If we use `\..\` in key, the app crashes and subsequent `readAll` calls also cause app crash.
  *  If we use `\..\..\` in key, `*.secure` file is saved outside of proper directory (it looks kinds of directry traversal valunurability).
* Keys length are limited to `MAX_PATH` - length of `%AppData%\{CompanyName}\{ProductName}`. It looks rare, but this restriction should be only Windows.
  * Currently, long key also causes app crash.
* Keys are case insenstive. It also looks rare, but this restriction should be only Windows.

### Environment

* `flutter_secure_storage` version: `8.0.0`

### Expected behavior

Even when we use in Windows,

* Keys can contain special chars including `:`.
* Safe to use keys conitains `..`.
* Keys are case sensitive as other platforms.
* Keys are case ssensitive.
* Throwing `PlatformException` rather than app crash when error.

### Proposal

First, I will appreciate for great works of current Windows implementation.

I have been tried to make PR for above, but I finally give up because I have to change the architecture to use encrypted JSON file (such as Linux implementation) for long key support but it is not so easy, and unfortunatelly, I found some small bugs including miss of resource clean-up and error handling.
So, I will propose that change Windows implementation to dart + `dart:ffi` from Visual C++ like [path_provider](https://github.com/flutter/packages/blob/a9ccbc2928633d9f0fe72590c922fa79080007bd/packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart). I think it is available and resonable because current implementation heavyly depends on Win32 APIs, dependency management (such as JSON handling) is more easy, unit testing becomes very easy including above edge cases. Any thoughts?

yfakariya:
Fixes #539

I changed the logic from files per keys to single encrypted JSON file to remove key restrictions which were introduced by NTFS specifications. And I also change encryption from BCrypt to [DPAPI](https://learn.microsoft.com/windows/win32/api/dpapi/) to avoid credential store cluttering.

* Remove restrictions:
  * Key length limit.
  * Key chars restriction.
  * No crash for `:/` pattern.
  * Key case sensitivity.
* This implementation uses DPAPI, so credential manager is not tainted
    with unreadable text.
* Full backward compatibility and auto-migration
* Options to disable backward compatibility and auto-migration.


juliansteenbakker:
Thank you very much for your contribution! It's a big rewrite, so i need some help from other people that can test this.

ccarpenter04:
It looks good to me, there are some additional options that could be added to make it more secure but as a baseline implementation it's better than what we have currently.

PR #549, opened 2023-04-13
[Linux] Remove libjsoncpp1 dependency

### LINKED ISSUE #433: Static linking of libjsoncpp
When I start my flutter application on a fresh Ubuntu install (ubuntu-22.04.1-desktop-amd64) I get the error:
> error while loading shared libraries: libjsoncpp.so.1: cannot open shared object file: No such file or directory

I found out that this comes from this dependency. Is there a way to static link this dependency so that I can escape the dependency hell?

Background is when I run this command:

>myApp/lib$ ldd libflutter_secure_storage_linux_plugin.so | grep json
>	libjsoncpp.so.1 => not found

The dependency is not found, if this would would be static linked into I would be able to run my app on linux with this flutter library.

Mino5531:
Removes the libjsoncpp1 dependency for Linux and replaces it with the a [header library](https://github.com/nlohmann/json)
This fixes #433 a dependency issue with Ubuntu-22.04+ and their derivatives, since libjsoncpp1 is no longer available on those distributions. 

Also has the added benefit, that it is no longer required by the end user to install libjsoncpp1 :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .buildlog/
 .history
 .svn/
+.vscode/
 migrate_working_dir/
 
 # IntelliJ related
--- a/flutter_secure_storage_linux/linux/CMakeLists.txt
+++ b/flutter_secure_storage_linux/linux/CMakeLists.txt
@@ -11,7 +11,6 @@ add_library(${PLUGIN_NAME} SHARED
 )
 apply_standard_settings(${PLUGIN_NAME})
 pkg_check_modules(LIBSECRET REQUIRED IMPORTED_TARGET libsecret-1>=0.18.4)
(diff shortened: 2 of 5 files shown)


spinus:
@Mino5531 how should security issues in library be applied here? Should be CI mechanism to detect updates in external library and create PR/notify about this? 
Usually if critical bug is found to library, linux distributions roll out the fix. There should be some scenario for that as it touches critical part of security system.

Mino5531:
Hi @spinus thanks for pointing that out. My goal was to make it possible to run apps using flutter_secure_storage on newer Ubuntu versions and to remove the need for the end user to install an additional library. I will try to add a way to link the library dynamically if the developer wishes to do so or create a CI mechanism (maybe some help with that, I'm not very experienced with CI), however that may take some time since I'm currently busy with other stuff.

juliansteenbakker:

PR #560, opened 2023-05-01
Fix compatibility with AGP 8.0

### LINKED ISSUE #557: Add compatibility with AGP 8 (Android Gradle Plugin)
Android Flutter plugins should add `namespace` under `android` inside the build configuration for the app.
Something like [this](https://github.com/fluttercommunity/plus_plugins/pull/1704/files).

Also check official android documentation [here](https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace).

### LINKED ISSUE #559: Android native plugin doesn't support AGP 8.0
The native Android plugin of this package doesn't support Android gradle Plugin 8.0, in 8.0 the package name of a library or application should be defined in build.gradle in the app module in the android block as a namespace instead of specifying it in the package attribute in AndroidManifest.xml, however, that might break compatibility with gradle versions older than 7.0, but gradle 8.0 will be required sooner or later, I think all the packages that have native plugin should update android gradle plugin to 8.0 so people start to migrate their build system to 8.0 to significantly improve the build system of android and how the app is packaged, All applications and libraries that use plugin have gradle version less than 8.0 can't migrate unless they update it manually but that is not a good solution since after any upgrade or clean the changes will reverted back

Also see [https://github.com/mogol/flutter_secure_storage/issues/557](https://github.com/mogol/flutter_secure_storage/issues/557)

[Fore more information](https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace)

Almighty-Alpaca:
See https://github.com/flutter/flutter/issues/125181 and https://github.com/flutter/flutter/issues/125621#issuecomment-1525995461 for more information

Fixes #559 and fixes #557

### DIFF
--- a/flutter_secure_storage/android/build.gradle
+++ b/flutter_secure_storage/android/build.gradle
@@ -22,6 +22,10 @@ allprojects {
 apply plugin: 'com.android.library'
 
 android {
+    if (project.android.hasProperty("namespace")) {
+        namespace("com.it_nomads.fluttersecurestorage")
+    }
+
     compileSdkVersion 33
 
     compileOptions {


CodeDoctorDE:
Looks like this repo is not very active?

Mr-Anderson-01 (approved):

Mr-Anderson-01:
@juliansteenbakker :)

juliansteenbakker:
Sorry for the very late response! LGTM!

josephcrowell:
Could you also push out an 8.x.x release which includes this fix?

pitw:
As as workaround until @juliansteenbakker  pushed the fixes to main branch import secure storage with this:

`  flutter_secure_storage:
    git:
      url: https://github.com/mogol/flutter_secure_storage.git
      ref: develop
      path: ./flutter_secure_storage`

incon:
I new release with this fix would be awesome :)

kev-techi:
@juliansteenbakker , a release of this would be greatly appreciated. Was forced to upgrade kotlin and gradle today. Can't compile Android without manually modifying this project's build.gradle file.

josephcrowell:
> @juliansteenbakker , a release of this would be greatly appreciated. Was forced to upgrade kotlin and gradle today. Can't compile Android without manually modifying this project's build.gradle file.

@pitw's fix works temporarily until julian pushes out a build with the fix.

kev-techi:
> > @juliansteenbakker , a release of this would be greatly appreciated. Was forced to upgrade kotlin and gradle today. Can't compile Android without manually modifying this project's build.gradle file.
> 
> @pitw's fix works temporarily until julian pushes out a build with the fix.

Sure, I just wouldn't recommend pointing to the develop branch for production apps as we don't know when something will merge and behavior could change.

Maybe the commit instead, so at least it stays stable?

```
git:
      url: https://github.com/mogol/flutter_secure_storage.git
      ref: bb3b811
      path: ./flutter_secure_storage
```

PR #570, opened 2023-05-25
[windows] Error when calling `delete()` before application support path exists

adg-mh:
Calling `delete()` while the application support path does not exist will throw an exception:

`PlatformException(Exception encountered: UNKNOWN_ERROR, delete, null, null)`

When calling DeleteFile() either ERROR_FILE_NOT_FOUND (0x2) or ERROR_PATH_NOT_FOUND (0x3) can be returned. This PR adds a check for ERROR_PATH_NOT_FOUND, as nothing should happen if the key does not exist.


juliansteenbakker:
LGTM!

PR #580, opened 2023-06-23
update js dep to 0.6.7

tim-eucalyptus:


juliansteenbakker:
Since this is a minor version, it is not needed to update it to the latest version in pubspec.yaml, since the pubspec.lock wil take care of this. This wil otherwise break for people that use an older version of this library.

PR #587, opened 2023-07-23
[Android] Get resetOnError from the options

CORDEA:
`resetOnError` is obtained from the `call.arguments` but this is in the options, so now I guess `resetOnError` will always be false.

In this PR, I updated it to get `resetOnError` from the options and also catch `deleteAll` errors to avoid unexpected crashes.


juliansteenbakker:

Issue #532, opened 2023-03-05
Flutter secure storage not storing data after I closed app IOS


stefanoHTB:
I am having problems storing a token with flutter secure storage!  

When I close the app or hot-restart the data gets lost or returns null

Any Suggestions? 

-------------------------------------------------------------------- code 

class Storage {
  final storage = const FlutterSecureStorage();

  final String _token = 'token';

  Future setToken(String token) async {
    const options =
        IOSOptions(accessibility: KeychainAccessibility.first_unlock);
    await storage.write(key: _token, value: token, iOptions: options);
  }

  Future<String?> getToken() async {
    const options =
        IOSOptions(accessibility: KeychainAccessibility.first_unlock);
    return await storage.read(key: _token, iOptions: options);
  }
}

BraveEvidence:
This will help https://www.youtube.com/watch?v=JogvnnaJ8TI

jannisnikoy:
I'm noticing this in a specific scenario where I log out and log in of my app. So basically, we wipe the stored data, and then store a new token when logging in. After restarting/hot restart;  same behavior as @stefanoHTB described. If I login and store a token after a clean start; there is no issue upon restart.

StigUK:
Hello, if you do getToken a second time, will the value still be null? I encountered a strange behavior that sometimes when I start the application, the value is returned null first time. However, it returns value after a certain amount of time when I try to do getToken again. And only on iOS.

jeff-odopass:
Please follow-up here by reacting to the original post, it is the same issue: https://github.com/mogol/flutter_secure_storage/issues/524

florentmorin:
Duplicates #524 ?

phillsch:
Having the same problem.

Did anyone found a solution?

codelovercc:
Having this same problem on MacOS. flutter secure storage version 9.0

rydmike:
@codelovercc Same issue on macOS, did not test iOS yet.
How are you all fixing this? I see no solution, at least not with an understandable explanation.

I'm beginning to think this PKG is not working well enough, pity. I will circle back to the issue later. I still need to test it on all platforms.

codelovercc:
@rydmike [Enable keychain](https://docs.amplify.aws/flutter/start/project-setup/platform-setup/#enable-keychain) this will help you. I've tested on macOS 12.7.1 (21G920) and Android Simulator API 34, it works.

rydmike:
[...]

Issue #519, opened 2023-02-08
Read on iOS fails silently and returns null instead of throwing platform exception.


janniklind:
Hi.

I am investigating this plugin a bit, since I want to store an encryption key for the local database in my app.
It is crucial that this key is not lost, otherwise the data in the database will be lost.

I am performing a standard check like (pseudo code):
```
key = flutterSecureStorage.read('key');
if (key == null) {
  // On first app start.
  flutterSecureStorage.write('key', cipher.generateKey());
}
database.open(key)
```

However it seems from the following line in your code, that you ignore the status on the response and always just returns the value:
https://github.com/mogol/flutter_secure_storage/blob/develop/flutter_secure_storage/ios/Classes/SwiftFlutterSecureStoragePlugin.swift#L47

Do I understand correctly that is fails silently and instead returns null? That would be bad for the most usage of this plugin since you cannot count on what you get from the read method.

gevork:
I having the same problem.
Wait any  solution

BraveEvidence:
This will help https://www.youtube.com/watch?v=JogvnnaJ8TI

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

PR #602, opened 2023-09-01
iOS Keychain - Use accessiblity option for all operations and throw PlatformException for SecErrors

### LINKED ISSUE #524: .read() return null since iOS 16.3.X even with the correct KeychainAccessibility
Package version: 

`flutter_secure_storage: ^8.0.0`

Flutter:

```
Flutter 3.3.10 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 135454af32 (vor 9 Wochen) • 2022-12-15 07:36:55 -0800
Engine • revision 3316dd8728
Tools • Dart 2.18.6 • DevTools 2.15.0

```

I have the following code: 

```dart
    aOptions: AndroidOptions(encryptedSharedPreferences: true),
    iOptions: IOSOptions(
      synchronizable: false,
      // https://support.apple.com/en-us/HT213407
      // https://stackoverflow.com/questions/10536859/ios-keychain-not-retrieving-values-from-background
      accessibility: KeychainAccessibility.first_unlock,
    ),
  );
```

```
  @override
  Future<String?> getAccessToken() async {
    final token = await _storage.read(key: "my_access_token");

   // The token is null, when the app was locked and we try to receive the token!! 😭

     return null;
  }
```

Regarding this question, the mode looks correct, but I just receive a null token when the app was moved into the background (not terminated) and the user locks the device. 
https://stackoverflow.com/questions/10536859/ios-keychain-not-retrieving-values-from-background

Thanks for the support.

bierbaumtim:
This PR fixes the issue that the accessiblity option was ignored for containtsKey and delete. Since all functions now require the conversion for the accessiblity option, I refactored that into a seperate function.

While working on this, I noticed that errors were not being propagated properly. So I improved the error propagation so that now errors will lead to a PlatformException. 

I also noticed that the implementation of containsKey could be optimized. Reading the value is not neccessary therefore I removed it and use the status to determine if the key is contained.

All changes currently apply to iOS only. Since the macOS implementation is almost identical, I could apply the fixes there too, but I can't test them because I don't have a Mac. What's your oppinion on that?

The work on this PR started as an attempt to fix #524. The changes does not fully fix the issue descripted in #524. To do fix them you need to wait until the function `applicationProtectedDataDidBecomeAvailable` is called and stop reading values when `applicationProtectedDataWillBecomeUnavailable` is called. I created a Gist containing the necessary code to do that.  https://gist.github.com/bierbaumtim/63c5057fc2889558d1801d95d93b6c6c
I think this plugin should support fetching and listening to changes in the availability of the keychain. But I want to know your opinion on that and if support should be added, should this be a seperate PR?


bierbaumtim:
@juliansteenbakker Could you review the changes?

juliansteenbakker:
Thank you for your work! Sorry for the late response, i'm planning on picking up some overdue work this plugin in the comming days. 

I think it will be good to add the same fixes on macOS. I can test them since i do have a mac. If you can create a separate PR for this, and for the new fetching and listening functions that would be great. Also, can you update this PR with the latest code base available?

bierbaumtim:
Thanks for your feedback. I plan to work on it on Saturday.

bierbaumtim:
PR is now update to date with latest code base

bierbaumtim:
Could also fix #532, #519

PR #617, opened 2023-09-25
Linux fix: Search with schemas fails in cold keyrings

inetic:
As reported in this issue: https://gitlab.gnome.org/GNOME/gnome-keyring/-/issues/89

If the keyring has not been previously unlocked (e.g. after restart of the PC) and if schema is used, invoking secret lookup will not show the password prompt to the user and libsecret will report the not finding the key.

Solution - as per the above issue - is to "warm up" the keyring by invoking the lookup function without a schema and discarding the result.

### DIFF
--- a/flutter_secure_storage_linux/linux/include/Secret.hpp
+++ b/flutter_secure_storage_linux/linux/include/Secret.hpp
@@ -10,12 +10,13 @@ class SecretStorage {
   FHashTable m_attributes;
   std::string label;
   SecretSchema the_schema;
+  bool cold_keyring;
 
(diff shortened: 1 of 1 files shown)


inetic:
Fixes https://github.com/mogol/flutter_secure_storage/issues/618

inetic:
Scratch that, there is still an issue with this approach: if during the warmup phase the user cancels, we'll not get an error and we'll assume warmup has completed successfully. The next lookup with `the_schema` will then again return nothing even if the entry is there.

Issue #618, opened 2023-09-25
Linux: Search with schemas fails in cold keyrings


inetic:
If the keyring has not been previously unlocked (e.g. after restart of the PC), invoking secret lookup will not show the password prompt to the user and the `read` method of `FlutterSecureStorage` will return `null`.

juliansteenbakker:
Fixed in #621

PR #621, opened 2023-09-26
Linux fix: Search with schemas fails in cold keyrings

inetic:
When the keyring is cold (e.g. after restart without a login or after running `gnome-keyring-daemon --replace` on the command line), lookup using schemas would return null entries without prompting the user for their passwords.

This is especially dangerrous in situations where the user wants to write an entry to the keyring. In such case, if there were previously stored secrets in the FlutterSecretStorage entry, those would be ignored and only the new entry would be stored. Effectively erasing all previously stored secrets.

The issue is explained here:
https://gitlab.gnome.org/GNOME/gnome-keyring/-/issues/89

However, the solution used in this commit is the one used in the chromium browser: http://crbug.com/660005


inetic:
Fixes https://github.com/mogol/flutter_secure_storage/issues/618

juliansteenbakker:

Issue #460, opened 2022-10-07
[Android] Stacktraces of exceptions should be reported via stacktrace property of PlatformException


ueman:
See this code

https://github.com/mogol/flutter_secure_storage/blob/1f773680b5bb237bb202f95858380ff4e38f992e/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java#L197-L199

The exception should be reported to the Dart side in accordance to the guidelines of the PlatformException.stackTrace property: https://api.flutter.dev/flutter/services/PlatformException/stacktrace.html

The benefit of doing that is that crash monitoring tools can then report the exception stacktrace properly

PR #623, opened 2023-09-29
Fix native Android stacktraces in PlatformExceptions

ueman:
Fixes https://github.com/mogol/flutter_secure_storage/issues/460

As described on the [`PlatformException.stacktrace`](https://api.flutter.dev/flutter/services/PlatformException/stacktrace.html) docs, native exceptions on Android don't need to be set via `result`. Instead, it has to be done by throwing a RuntimeException.

If an exception isn't propagated to Flutter as described, the `PlatformException.stacktrace` property isn't populated. This negatively impacts tooling like Sentry, and other error monitoring tools.

Let me know if there are any follow-up changes are known.


juliansteenbakker:
I did not know this yet, thanks for the contribution!

Issue #581, opened 2023-06-25
Write on Mac doesn't overwrite previous value; Delete works fine;


tomekit:
I am experiencing this issue only on macOS.
I suspect it's something related to Keychain, but I don't yet have enough experience with the lib internals to trace exactly what's going on.

In some cases when I write new value to a `key`, the new value doesn't end up in: `key`. It's like if the write to keychain has failed.
However when I delete the: `key` it gets deleted as expected.

It doesn't make difference if I try to read new value immediately after write or if I wait longer while or even restart app.
 
```dart
final secureStorage = FlutterSecureStorage();
final write = await secureStorage.write(key: "key", value: "newValue");
final value = await secureStorage.read(key: "key"); // Still returns old value !

final delete = await secureStorage.delete(key: "key");
final value = await secureStorage.read(key: "key"); // Key successfully deleted
```

I use latest: `flutter_secure_storage 8.0.0` and Flutter:
```
[✓] Flutter (Channel stable, 3.10.5, on macOS 13.4 22F66 darwin-arm64, locale en-PL)
    • Flutter version 3.10.5 on channel stable at /Users/tomek/development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 796c8ef792 (12 days ago), 2023-06-13 15:51:02 -0700
    • Engine revision 45f6e00911
    • Dart version 3.0.5
    • DevTools version 2.23.1
```

This issue doesn't seem to be present on Linux, Windows, iOS, Android and Web.

==

I've also noticed strange inconsistency between key versions depending if I run the "Debug" version of the app from the Android Studio or if I start the "Release" version from the compiled .DMG package.

Once I start release version of the app, I get this prompt
![signal-2023-06-25-220605_002](https://github.com/mogol/flutter_secure_storage/assets/7759113/a2aba64b-2e0a-4810-a528-6f8a99e1b7e9)
and the if I read the `key` I get the: `newValue`.

After closing the app and starting the Debug version from the Android Studio:
![signal-2023-06-25-220605_003](https://github.com/mogol/flutter_secure_storage/assets/7759113/ed1e21af-8539-4072-9fef-2cd564eaed34)
I still get the: `oldValue`, despite the fact the the: `newValue` was written from the Debug version.

clong1995:
I had the same problem

SebAubin:
Same, we have to delete it before writing the data again.

cst1412:
[...]

Issue #350, opened 2021-12-28
Mac OS write read doesn't work


jaumard:
Not sure why but on mac OS nothing is saved (or read) from secure prefs :(

```dart
final securePrefs = const FlutterSecureStorage();
await securePrefs.write(key: 'keyToken', value: 'token');
final token = await securePrefs.read(key: 'keyToken');
```
token is always null, no errors or anything. No problems on iOS and Android

Using `flutter_secure_storage: ^5.0.2` and `flutter 2.5.3`

JWambaugh:
I'm having the same issue :/. No errors given, just doesn't work on mac.

Andrflor:
I have the exact same problem on latest ArchLinux, always null, no error :face_with_spiral_eyes:

zebamba:
I am Using flutter_secure_storage: ^5.0.2 and flutter 2.9.0-1.0.pre.471

I have same issue on Mac OS.

Is there any required permission to set/change while the app is in development?

mikexkllr:
I have the same issue. Is there any configuration which needs to be done in the MacOS runner? thx for help!

zebamba:
I opened the project on XCode, then added Keychain Sharing. Added group value as * (asterisk symbol) 

I am Using flutter_secure_storage: ^5.0.2, flutter 2.10.4, Dart 2.16.2

The same error still happening returning null value from the storage. 

What I am missing here?

<img width="606" alt="Screen Shot 2022-04-08 at 12 41 00 PM" src="https://user-images.githubusercontent.com/10945353/162486834-05dce1eb-79fe-41ae-bb01-cbdd946aa712.png">

<img width="699" alt="Screen Shot 2022-04-08 at 12 42 19 PM" src="https://user-images.githubusercontent.com/10945353/162486898-ccb60d57-2d7d-4904-9a5e-46473ce69e07.png">



<img width="875" alt="Screen Shot 2022-04-08 at 12 42 43 PM" src="https://user-images.githubusercontent.com/10945353/162486175-159307f4-4d39-4f1c-b653-9238953d3268.png">

fluttertests:
same thing here and brought me here.

zebamba:
@fluttertests @mikexkllr @jaumard @Andrflor 

I got it working, but I have few other issues with the build for mac after changes on the project.

Try this.

Open the `macos` project folder on Xcode
Add Key Chain sharing capability

[...]

Issue #302, opened 2021-09-02
macOS - containsKey is always returning true


kineticac:
No matter what key I'm passing for the `containsKey` function, this always returns true the first time on a macOS build.

Is anyone else getting this issue? Let me know what other information you might need to help resolve.

pikawika:
I have the same issue on iOS Iphone 12 simulator -> containsKey returns true but read returns null, used to work on stable release. temporary bypass for me is a function to check contains by calling
`await _secureStorage.containsKey(key: _tokenKey) && await _secureStorage.read(key: _tokenKey) != null;`

juliansteenbakker:
Please try again using the latest v5.0.0. If issue still exists feel free to comment.

AgustinV08:
Problem still exists in package 5.0.2

jscti:
+1

Pretty critical/hardcore bug when you use this function to check if a user is LoggedIn or not ...

Maybe this plugin is not so secure ^^

juliansteenbakker:
> Pretty critical/hardcore bug when you use this function to check if a user is LoggedIn or not ...

You should not use this function to check if a user is logged in or not. You should read the actual data that it has and check if it is still valid or not expired like a token.

jscti:
It's up to me and you don't know of any of the other technical details. If you think containsKey has no use, remove it from the lib

But yeah now that containsKey is totally broke up, I don't really have a choice

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

PR #628, opened 2023-10-06
macOS Keychain - Use accessiblity option for all operations and throw PlatformException for SecErrors

bierbaumtim:
Changes from PR #602 but for macOS.


bierbaumtim:
Could possibly fix or at least fail with a native error message #581, #350 and #302.

juliansteenbakker:
Tested and working, thanks again !

Issue #595, opened 2023-08-10
Keychain data on iOS gets deleted after some time


ilyosbekkk:
Hi, thank you for that amazing package! I am having a strange issue in my app, I store all tokens in secure storage in my Flutter app and it's working perfectly for most  of the users but one of my users didn't use the app actively for weeks (or even a month) and app was either in background or terminated state. Today he opened the app and noticed that all requests to the server didn't work.  We found out that when we try to get the token from secure storage it returns `null`.  What might be the reason for this? Does the iOS delete the data from Keychain if it's not active for some time? Thank you for that answer in advance:)

Fellerson:
Hello
Any update about this issue ? We have the same problem by us. Does somebody fixed it ? ty

OneOFFER:
Hi 
We have the same problem. Does soomebody fixed it ?
Thank you

Fellerson:
Hi,
It looks that nobody fixed it. We tried to reach them by mail unsuccessfully.

KanybekMomukeyev:
Same here

roncakjakub:
same here :)

neiljaywarner:
see the code in develop re: cupertinoProtected data availability from above mr

g-apparence:
Is there any fix? 
This makes the all plugin unusable in production

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

juliansteenbakker:
Some extra information for this issue, there is a known issue in version 9.1.0 which causes data to be inaccessible due to wrong read parameters. This will be fixed in the upcoming release.

Issue #622, opened 2023-09-28
Flutter Secure Storage Data is lost sometimes in both Android and IOS, what should I do not to lost automatically in secure storage.


KyawLwinSoe27:

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

Issue #614, opened 2023-09-15
Weird issue where for iOS only


curtis2101:
There is an issue where the top line of a storeAndOverwrite function is saved and able to be read when the app is running. But on restart of the app the top line that was written could not be found at all. This happens only on IOS and to fix this I Have added a future delay for now. So in this example without a delay the biometric preference would read while the app is running and its just been saved but on restart it would not show but all the other keys and values would be.
Future<void> storeAndOverwrite(SecureStorageContent storedContent) async {
    await Future.delayed(const Duration(
        milliseconds:
            200)); //do not remove this line seems to be a ios issue where the top line is not stored
    await storage.write(
        key: Dummy.biometricPreference,
        value: storedContent.biometricPreference);
    await storage.write(
        key: Dummy.soundPreference,
        value: storedContent.soundPreference);
    await storage.write(
        key: Dummy.vibrationPreference,
        value: storedContent.vibrationPreference);
    await storage.write(key: LegacyConstants.email, value: storedContent.email);
    await storage.write(
        key: Dummy.password, value: storedContent.password);
    await storage.write(
        key: Dummy.accessToken, value: storedContent.accessToken);
    await getStoredContent();
  }

harshitwe360:
I have updated with new IOS versions, after that I had got same issue, String type data is not get stored. but its working fine in android, even in IOS debug mode not working in release mode. 

Any idea whats going wrong..!

MatthewJones517:
I am having the same issue with 9.0.0. Took me forever to track down that this plugin was the problem. Issue only occurs in iOS release builds. Nowhere else.

MatthewJones517:
After digging in a bit deeper, I found that this was caused by doing a read immediately after doing a write. We weren't fully awaiting the save, so only part of the value made it in to secure storage before we pulled it out. 

I'm not saying that's everyone's problem, but it's something to check for those that stumble on this thread in the future.

Kijacode:
so @MatthewJones517  , have you solve it ??

cheymos:
Is the problem still relevant?

aswinranjith-iouring:
any update on this?

juliansteenbakker:
Version 9.2.1 fixed a race condition issue with read/write, which could be this issue. It would be great if someone can check if this version still has this issue.

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

PR #629, opened 2023-10-07
Feature - Check availability of protected data on iOS and macOS

### LINKED ISSUE #524: .read() return null since iOS 16.3.X even with the correct KeychainAccessibility

bierbaumtim:
As discussed in PR #602 this PR adds the bridge to `isProtectedDataAvailable` on macOS(12.0 and newer) and iOS. It also adds the possiblity to get notified when availability of the protected data will change. At the moment listening to `applicationProtectedDataDidBecomeAvailable` and `applicationProtectedDataWillBecomeUnavailable` is only possible on iOS. Support for macOS is currently blocked by PR https://github.com/flutter/engine/pull/44587 not beeing in the current flutter stable version.
As soon as a new major stable version of flutter gets released we could add support for macOS.

Combined with PR #602 this fixes #524.


bierbaumtim:
Could fix #595, #614, #622, #532

flutter_secure_storage/ios/Classes/SwiftFlutterSecureStoragePlugin.swift
```diff
@@ -6,15 +6,20 @@
...
-public class SwiftFlutterSecureStoragePlugin: NSObject, FlutterPlugin {
+public class SwiftFlutterSecureStoragePlugin: NSObject, FlutterPlugin, FlutterStreamHandler {
     
     private let flutterSecureStorageManager: FlutterSecureStorage = FlutterSecureStorage()
-    
+    private var secStoreAvailabilitySink: FlutterEventSink?
+
     public static func register(with registrar: FlutterPluginRegistrar) {
         let channel = FlutterMethodChannel(name: "plugins.it_nomads.com/flutter_secure_storage", binaryMessenger: registrar.messenger())
+        let eventChannel = FlutterEventChannel("plugins.it_nomads.com/flutter_secure_storage/events", binaryMessenger: registrar.messenger())
```
gnunicorn, line 18: This fails for me on the swift compiler, because it is missing the `name:` prefix. This fixes it:

```suggestion
        let eventChannel = FlutterEventChannel(name: "plugins.it_nomads.com/flutter_secure_storage/events", binaryMessenger: registrar.messenger())
```

bierbaumtim: Oh yeah right. Thanks for the fix.

juliansteenbakker:
Looks good to me, thanks for your work!

bogdannedelcu:
Is this available in the public package version 9.0.0?

neiljaywarner:
@bogdannedelcu  looks like it isn't right?

PR #640, opened 2023-10-20
Listener functionality

MohamadGreatWarrior:
- Each listener will be called with the value of the respective key injected as argument.
- You can register as many listeners as you want for each key. 
- When you delete a key its listeners will still be called with `null` injected as the value for the listener.

- The ability to un-register one listener for specific key with keeping other listeners for that same key.
- The ability to un-register all listeners for specific key.
- The ability to un-register all listeners for all keys.


juliansteenbakker:
Great! LGTM!

PR #641, opened 2023-10-25
Implement isCupertinoProtectedDataAvailable and onCupertinoProtectedDataAvailabilityChanged for web platform

hoangdat:


flutter_secure_storage_web/lib/flutter_secure_storage_web.dart
```diff
@@ -181,4 +181,11 @@ class FlutterSecureStorageWeb extends FlutterSecureStoragePlatform {
 
     return plainText;
   }
+
+  @override
+  Future<bool> isCupertinoProtectedDataAvailable() => Future.value(true);
+
+  @override
+  Stream<bool> get onCupertinoProtectedDataAvailabilityChanged =>
+      Stream.value(true);
```
tenhobi, line 190: I completely agree this should be implemented in some way. Hovewer should this be `true` for web? I think it should return `false` for the  `isCupertinoProtectedDataAvailable` and `Stream.empty()` (I think there is something like that?) for `onCupertinoProtectedDataAvailabilityChanged`.

tenhobi:
Guys this should have been fixed since without that the package throws an error on build...

PR #662, opened 2023-12-10
Resolve error when no item exist (iOS)

### LINKED ISSUE #391: Data of FlutterSecureStorage instance is null on iOS but not on Android
Hello, this code works on my Android app but not iOS. I tested it on a physical device and simulator. 

```
class _MyHomePageState extends State<MyHomePage> {
  final storage = const FlutterSecureStorage();

  Future _containsKey() async {
    return storage.containsKey(key: "conciergeRowGuid");
  }

  @override
  Widget build(BuildContext context) {
    return FutureBuilder(
        future: _containsKey(),
        builder: (context, snapshot) {
          return (snapshot.data == false)
              ? Scaffold(
                  backgroundColor: Colors.white,
                  body: Container(child: Text("not loaded"))
                )
              : FutureBuilder(
                  future: storage.read(key: "conciergeRowGuid"),
                  builder: (context, snapshot) {
                    if (snapshot.data == null) {
                      return SizedBox(
                        height: MediaQuery.of(context).size.height - 135,
                        width: MediaQuery.of(context).size.width,
                        child: Column(
                          mainAxisAlignment: MainAxisAlignment.center,
                          crossAxisAlignment: CrossAxisAlignment.center,
                          children: const [
                            CircularProgressIndicator(
                              color: Color.fromARGB(255, 176, 64, 251),
                            ),
                          ],
                        ),
                      );
                    }
                    return Container(child: Text("loaded"));
                  });
        });
  }
```

In the iOS app, `snapshot.data` is always null. Digging into the iOS code of the Flutter Secure Storage, the functions of `containsKey`, `write`, `read`, `readAll`, `deleteAll`, all have `groupId` as `nil`. Also, `OSStatus status` has the error code 25300 which means `errSecItemNotFound `. And the return `NSString *value` is also `nil`. There is a similar issue #107 but the fix suggested is not working for me. 

Specs:
Flutter 2.10.4
`flutter_secure_storage: ^5.0.2`
MacOS Big Sur 11.6
Tested on:
- iPhone 13 Simulator iOS 15.0
- Physical iPhone 11 iOS 14.6
- Physical iPhone SE (2nd gen) iOS 15.4.1

koji-1009:
Fixes #391

On iOS, `readAll` (or `deleteAll`) will return error code `errSecItemNotFound`(`-25300`) if the item does not exist in the keychain. This is due to the error code `errSecItemNotFound` returned by `SecItemCopyMatching` (or `SecItemDelete`).
This is expected behavior for iOS processing and is not a bug.

https://developer.apple.com/documentation/security/1542001-security_framework_result_codes/errsecitemnotfound

Therefore, this fix modifies the functions to return `nil` in cases when `errSecItemNotFound` is returned.

### DIFF
--- a/flutter_secure_storage/example/ios/Podfile.lock
+++ b/flutter_secure_storage/example/ios/Podfile.lock
@@ -2,21 +2,28 @@ PODS:
   - Flutter (1.0.0)
   - flutter_secure_storage (6.0.0):
     - Flutter
+  - path_provider_foundation (0.0.1):
+    - Flutter
+    - FlutterMacOS
 
 DEPENDENCIES:
   - Flutter (from `Flutter`)
   - flutter_secure_storage (from `.symlinks/plugins/flutter_secure_storage/ios`)
+  - path_provider_foundation (from `.symlinks/plugins/path_provider_foundation/darwin`)
 
 EXTERNAL SOURCES:
   Flutter:
     :path: Flutter
   flutter_secure_storage:
     :path: ".symlinks/plugins/flutter_secure_storage/ios"
+  path_provider_foundation:
+    :path: ".symlinks/plugins/path_provider_foundation/darwin"
 
 SPEC CHECKSUMS:
   Flutter: f04841e97a9d0b0a8025694d0796dd46242b2854
   flutter_secure_storage: 23fc622d89d073675f2eaa109381aefbcf5a49be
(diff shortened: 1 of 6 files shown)


tenhobi:
We would love to see merged since we also get `errSecItemNotFound` instead of `null`. 👍

juliansteenbakker:
Looks good, thanks for your contribution!

koji-1009:
@juliansteenbakker 
Thanks!

PR #673, opened 2024-01-24
[LINUX] Fix erase was called on null #616

### LINKED ISSUE #616: App fails with: nlohmann::json_abi_v3_11_2::detail::type_error
We're affected by the `libjson` replacement with https://github.com/nlohmann/json in this commit: https://github.com/mogol/flutter_secure_storage/commit/647a5d1035444959dc9c3034a78a13154493b2dc 

Below code
```dart
final random = Random();
final intRandom = random.nextInt(100000000);
final secureStorage = SecureStorage();
await secureStorage.write(key: 'testKey', value: intRandom.toString());
final readInt = await secureStorage.read(key: 'testKey');
context.showSnackBar(message: "Generated: ${intRandom.toString()}; Read after save: ${readInt.toString()}");
```
crashes the app, with below error:
```
terminate called after throwing an instance of 'nlohmann::json_abi_v3_11_2::detail::type_error'
  what():  [json.exception.type_error.307] cannot use erase() with null
```

We can't reproduce it on the machine that we compile the code on, however once we pass code to another machine, e.g. using Flathub format: https://github.com/flathub/io.kapsa.drive/blob/master/io.kapsa.drive.json#L252 we get this error.

Our solution was to revert this commit in our fork: https://github.com/Mino5531/flutter_secure_storage/compare/develop...tomekit:flutter_secure_storage:develop

and load it like this:
```yaml
dependencies:
  flutter_secure_storage:
    git:
      url: https://github.com/tomekit/flutter_secure_storage.git
      ref: develop
      path: flutter_secure_storage
dependency_overrides:
  flutter_secure_storage_linux:
    git:
      url: https://github.com/tomekit/flutter_secure_storage.git
      ref: develop
      path: flutter_secure_storage_linux
```

Do you have an idea why this error may occur on some machines?

Out of curiosity what was the reason of moving away from `libjsoncpp25`?

prateekmedia:
Try Catched the exception, it basically happens when key is not present in libsecret.

fix #616

### DIFF
--- a/flutter_secure_storage_linux/linux/include/Secret.hpp
+++ b/flutter_secure_storage_linux/linux/include/Secret.hpp
@@ -45,12 +45,16 @@ public:
   }
 
   void deleteItem(const char *key) {
-    nlohmann::json root = readFromKeyring();
-    if (root.is_null()) {
-        return;
+    try {
+      nlohmann::json root = readFromKeyring();
+      if (root.is_null()) {
+          return;
+      }
+      root.erase(key);
+      storeToKeyring(root);
+    } catch (const std::exception& e) {
+        std::cerr << "Exception: " << e.what() << std::endl;
     }
-    root.erase(key);
-    storeToKeyring(root);
   }
 
   bool deleteKeyring() { return this->storeToKeyring(nlohmann::json()); }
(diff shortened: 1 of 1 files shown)


prateekmedia:
@mogol The workflows doesn't seem to be due to my PR as it doesn't disturb integration_test on android nor does it cause describeEnum error in code analysis.

Can you review the PR?

juliansteenbakker:
Sorry for the late response, the code looks good! LGTM!

PR #680, opened 2024-02-17
Update js dependency and migrate to web

ThexXTURBOXx:
This allows newer versions of `js` being used in conjunction with `flutter_secure_storage`. Also, as `html` is about to get phased out, this PR already migrates to `web` as it is recommended practice to do so.

Fixes #670 
Fixes #679 
Fixes #683

If you want to use this **now**, you can do so via a `dependency_override`:
```yaml
dependency_overrides:
  flutter_secure_storage:
    git:
      url: https://github.com/ThexXTURBOXx/flutter_secure_storage.git
      ref: develop
      path: flutter_secure_storage
```
You probably need to also override the other subpackages similarly (I won't explain how to do that. Just do it as I have shown with the main package above).

PR #691, opened 2024-03-14
chore: upgrade js from 0.6.3 to 0.7.1

narimanav115:


juliansteenbakker:
Hi all, i am working on integrating #680, so this should not be needed anymore. Thank you anyway!

PR #693, opened 2024-03-27
updated js

Chimela-roar:
I update the js from 0.6.3 to 0.7.1 because i tried adding a package from on my app but i wasn't able to update the pubspec.ymal because the js wasn't up to date


juliansteenbakker:
Hi all, i am working on integrating #680, so this should not be needed anymore. Thank you anyway!

PR #710, opened 2024-05-09
Return nil on iOS read if key is not found

### LINKED ISSUE #709: The specified item could not be found in the keychain
Hello,
after yesterday's update (9.1.1) we are encountering this error when trying to read secure storage only on ios:

`Unhandled Exception: PlatformException(Unexpected security result code, Code: -25300, Message: The specified item could not be found in the keychain., -25300, null)`

How can we solve that?
Thank you

ztaylor54:
Fixes #709.

It looks like the `readAll` function was [updated in a recent commit](https://github.com/mogol/flutter_secure_storage/commit/a2ba4673842e5fe840f6928c8a4d9481060d0dbd#diff-a18a015af5fc9438706d6f278f142404f290313db58898aae4b3df63f9e8e598R86-R90) to explicitly return `nil` when a keychain key isn't found, but the `read` function was not updated to match the new error checking scheme.

I just added the same check for `errSecItemNotFound` and I am able to get my builds working on version 9.1.1.

### DIFF
--- a/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
+++ b/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
@@ -139,6 +139,11 @@ class FlutterSecureStorage{
     internal func delete(key: String, groupId: String?, accountName: String?, synchronizable: Bool?, accessibility: String?) -> FlutterSecureStorageResponse {
         let keychainQuery = baseQuery(key: key, groupId: groupId, accountName: accountName, synchronizable: synchronizable, accessibility: accessibility, returnData: true)
         let status = SecItemDelete(keychainQuery as CFDictionary)
+
+        // Return nil if the key is not found
+        if (status == errSecItemNotFound) {
+            return FlutterSecureStorageResponse(status: errSecSuccess, value: nil)
+        }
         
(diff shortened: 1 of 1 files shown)


ztaylor54:
@paulking86 I think you're right here -- reading through recent comments in #709 shows that the issue is with more than just the `read` function. I should have time to update this PR today.

ztaylor54:
@juliansteenbakker I also added the check to the `delete` function, as it is also affected. 

[Someone mentioned](https://github.com/mogol/flutter_secure_storage/issues/709#issuecomment-2106046427) that `write` was also affected, but I am unsure how to reproduce so I'll leave that for you to triage.

I suspect this has to do with a change to keychain in recent iOS versions. The app store requirements recently changed to require iOS 17 SDK for all submissions, but I can't find anything in Apple's changelogs that points to an obvious culprit.

juliansteenbakker:
Looks good, LGTM!

aliasgar4558:
@juliansteenbakker - Are we not looking for package version update for this fix?

juliansteenbakker:
I am still working on some other fixes for macOS and iOS, so the update will be later today

PR #712, opened 2024-05-13
Also set kSecUseDataProtectionKeychain on read for macos

gnunicorn:
We were experiencing problems when using `synchronizable: false` on our Mac setup. After some investigation I found out that despite successfully able to write, even immediately after, the read would not return the updated value. Turns out that setting [`kSecUseDataProtectionKeychain`](https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain) on write also means we need to set that on read or the value isn't properly found. The docs also recommend setting it to true always and we are doing that on write already, but weren't on the query.

This patch sets that key to true regardless of synchronizable being set fixing the issue for us.


juliansteenbakker:

PR #717, opened 2024-05-15
Fix async race condition bug in storage operations

### LINKED ISSUE #716: [bug] Race condition on `write` and `read`.
## Problem
Race condition on `write` and `read`.

There is `FlutterSecureStorage` operations `async` bug:

``` dart
Future<void> main() async {
  WidgetsFlutterBinding.ensureInitialized();
  const storage = FlutterSecureStorage();
  await storage.write(key: 'key', value: 'value');
  assert(await storage.read(key: 'key') == 'value'); // assertion: platform [read] finishes before platform implementation [write].
}
```

PROGrand:
## Description
Race condition on `write` and `read`.

- [ ] Closes #716

## Solution
Using of `await` is necessary when switching from expression body returning feature (`=>`) to async block body (`{}`).

### DIFF
--- a/flutter_secure_storage/CHANGELOG.md
+++ b/flutter_secure_storage/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 9.2.1
+* Fix async race condition bug in storage operations.
+
 ## 9.2.0
 New Features:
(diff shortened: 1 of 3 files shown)


juliansteenbakker:

PR #730, opened 2024-06-01
feat: Support for wrap and unwrap key on web platform

### LINKED ISSUE #726: [Web] Encryption of saved value by app-specific key
The current implementation is "secure" in the sense that when a user looks at LocalStorage, they cannot immediately identify the value. (I think that's enough for most cases.)

https://github.com/mogol/flutter_secure_storage/blob/v9.2.2/flutter_secure_storage_web/lib/flutter_secure_storage_web.dart#L104

On the other hand, by reading the code in flutter_secure_storage, we can analyze the stored jwk. (Of course, it is hard work.)
To improve this analysis difficulty, I propose to obfuscate the stored jwk using an app-specific key.

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/wrapKey
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/unwrapKey

How about adding this option to `WebOptions`?

koji-1009:
close #726

I have added a function to wrap a key stored in LocalStorage using an application specific encryption key. This function uses the WebCrypto API [wrapKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/wrapKey) and [unwrapKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/unwrapKey).

The wrapping key can be generated by the following snippet.

```javascript
async function main() {
  const iv = new Uint8Array(12);
  window.crypto.getRandomValues(iv);

  const key = await window.crypto.subtle.generateKey(
    {
      name: "AES-GCM",
      length: 256,
      iv: iv,
    },
    true,
    ["wrapKey", "unwrapKey"]
  );

  const jsonWebKeyBuffer = await window.crypto.subtle.exportKey("raw", key);
  const jsonWebKey = new Uint8Array(jsonWebKeyBuffer);

  console.log("---iv---");
  const base64Iv = btoa(String.fromCharCode.apply(null, iv));
  console.log(base64Iv);

  console.log("---wrapping key---");
  const wrappingKey = btoa(String.fromCharCode.apply(null, jsonWebKey));
  console.log(wrappingKey);
}

main();
```

To use the function, set a value to the option as follows.

```dart
  // This is example key and iv
  webOptions: WebOptions(
    wrapKey: 'FQQloR/Xwv1BcJOc3VrcxlJatdA/C+n6+v7pFJC0GBs=',
    wrapKeyIv: 'YTVC7K5dOsQTjk+z',
  ),
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,22 @@ Please see:
 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
 - https://www.netsparker.com/blog/web-security/http-security-headers/
 
+#### application-specific key option
+
+On the web, all keys are stored in LocalStorage. flutter_secure_storage has an option for the web to wrap this stored key with an application-specific key to make it more difficult to analyze.
+
+```dart
+final _storage = const FlutterSecureStorage(
+  webOptions: WebOptions(
+    wrapKey: '${your_application_specific_key}',
+    wrapKeyIv: '${your_application_specific_iv}',
+  ),
+);
+```
+
+This option encrypts the key stored in LocalStorage with [WebCrypto wrapKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/wrapKey). It is decrypted with [WebCrypto unwrapKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/unwrapKey) when used.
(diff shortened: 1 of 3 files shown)


juliansteenbakker:
Thank you for this function! It would be great if we have some documentation on how to implement this feature in the readme

koji-1009:
@juliansteenbakker 
Thanks for the review! I have reflected the changes in the README.

juliansteenbakker:
Thank you ! LGTM!

PR #731, opened 2024-06-01
feat: Improve deleteAll in FlutterSecureStorageWeb

### LINKED ISSUE #619: [web] `deleteAll` clears entire local storage
https://github.com/mogol/flutter_secure_storage/blob/57c2b54c3e1a839d1f6754a487b85b44e027b46d/flutter_secure_storage_web/lib/flutter_secure_storage_web.dart#L49

This line will remove *everything* in local storage, even if they were inserted by other libraries or the App.

Suggestions:
1. only clear the keys that were inserted through secure storage
2. or, only clear the keys that are prefixed with `${options[_publicKey]!}.`. It's a little riskier but saves memory.
What do you think?

koji-1009:
close #619

Fixed `deleteAll` function to only delete values added by flutter_secure_storage_web.


juliansteenbakker:
Looks good, thanks!

PR #719, opened 2024-05-20
fix: [iOS / macOS] pass accessibility to `readAll` and `deleteAll`

techouse:
As described in https://github.com/mogol/flutter_secure_storage/issues/709#issuecomment-2114743623 I was having issues getting values from my keychain using `await secureStorage.readAll()`, however, `await secureStorage.read(key: 'foo')` worked fine. Additionally, @feinstein reported similar issues with `deleteAll` in #720 

It seems that somehow the `accessibility` for `readAll` and `deleteAll` was explicitly (possibly deliberately?) set to `nil`  in #602 

- ### iOS
  - `accessibility` is not passed in `readAll` https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift#L75 whereas it's passed in `read` https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift#L105
  - `deleteAll` is also affected (reported in #720) https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift#L128 opposed to `delete` https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift#L140

- ### macOS
  - The same is true for the `macOS` package's `readAll`
https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage_macos/macos/Classes/FlutterSecureStorage.swift#L78 vs `read` https://github.com/mogol/flutter_secure_storage/blob/eaccd9fb88effab9d3274c51cc91a3bc5886d8fd/flutter_secure_storage_macos/macos/Classes/FlutterSecureStorage.swift#L108
[...]

Issue #711, opened 2024-05-11
Message: The specified item already exists in the keychain, IOS 


yourior:
> [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 
> PlatformException(Unexpected security result code, Code: -25299, Message: The specified item already exists in the keychain., -25299, null)
> #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:651:7)
> #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:334:18)
> <asynchronous suspension>
> #2      WriteStringStorage (package:pkgname/Storage/Secure_Storage.dart:34:10)
> <asynchronous suspension>
> #3      GoRouterSplashScreen.build.<anonymous closure>.<anonymous closure> (package:pkgname/Splash_Screen.dart:39:13)
> <asynchronous suspension>
> #4      GoRouterSplashScreen.build.<anonymous closure> (package:pkgname/Splash_Screen.dart:36:9)
> <asynchronous suspension>
> #5      _FlutterSplashScreenState.didChangeDependencies (package:another_flutter_splash_screen/another_flutter_splash_screen.dart:325:7)
> <asynchronous suspension>
> Type: Notice | Timestamp: 2024-05-11 09:26:17.155593+07:00 | Process: Runner | Library: Flutter | TID: 0x12e46f

here is my initialize code 
```
AndroidOptions _getAndroidOptions() => const AndroidOptions(
  encryptedSharedPreferences: true,
);

IOSOptions _getIOSOptions() => const IOSOptions(
  // synchronizable: true,
  accessibility: KeychainAccessibility.first_unlock
);
final storage = FlutterSecureStorage(
  aOptions: _getAndroidOptions(),
  iOptions: _getIOSOptions()
);
```
the error showed when i want to write a value into a key

`storage.write(key: Keyname,value: value);`

>  version: "9.1.1"

is this a bug or is there a setting that i need to set for IOS? because this works in android

KasperTidemann:
@yourior I’m seeing the same error here, not sure what it’s related to just yet. It used to work just fine, so I’m currently investigating.

yourior:
@KasperTidemann it works by initializing it using the default method
`final storage = const FlutterSecureStorage();`
then call the option using the function
`storage.write(key: Keyname,value: value,iOptions: _getIOSOptions(),aOptions: _getAndroidOptions());`

[...]

Issue #692, opened 2024-03-24
[iOS] Native app update


pawelballo99:
I'm trying to get keychain data from native app after upadating it with flutter app. Flutter app has bundleID and accountName for keychain the same as native app, but `final data = await FlutterSecureStorage().readAll();` is always empty.

SERDUN:
Hi, I have the same issue after updating the plugin from 
flutter_secure_storage: 9.0.0 
to 
flutter_secure_storage: 9.2.1


`Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.22.0, on macOS 14.4.1 23E224 darwin-arm64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2023.3)
[✓] Connected device (4 available)
[✓] Network resources

• No issues found!
`

Try calling .readAll() immediately after the WidgetsFlutterBinding.ensureInitialized(); and before runApp(...).


<img width="653" alt="Screenshot 2024-05-19 at 12 36 06" src="https://github.com/mogol/flutter_secure_storage/assets/26858237/4a088db8-44af-4c31-9b31-00a7b2d3a6e1">

techouse:
This was fixed in #719

sureshprokraya:
Hi, 
I am using flutter_secure_storage: 9.0.0 version in flutter application. But when ever i closed the app and reopen again , 
After splash screen  i am getting data null in IOS, but in Android works fine. 

How to fix this issue,  please any one help me on this.

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

PR #734, opened 2024-06-07
refactor: Cleanup optional values

koji-1009:
https://github.com/mogol/flutter_secure_storage/blob/v9.2.2/flutter_secure_storage/lib/options/apple_options.dart
https://github.com/mogol/flutter_secure_storage/blob/v9.2.2/flutter_secure_storage/lib/options/ios_options.dart

In AppleOptions and IOSOptions, `synchronizable` and `accessibility` are treated as non-null values. Also, since it is defined as `Map<String, String>` in Dart, it can be treated as `[String : String]` in Swift.
Because of the above, the `synchronizable` and `accessibility` properties of the methods are non-nil. Because of the above, the code will not be able to use the `?` has been removed.

### DIFF
--- a/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
+++ b/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
@@ -8,11 +8,7 @@
 import Foundation
 
 class FlutterSecureStorage{
-    private func parseAccessibleAttr(accessibility: String?) -> CFString {
-        guard let accessibility = accessibility else {
-            return kSecAttrAccessibleWhenUnlocked
-        }
-        
+    private func parseAccessibleAttr(accessibility: String) -> CFString {        
         switch accessibility {
         case "passcode":
             return kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly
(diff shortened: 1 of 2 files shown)


dballance:
This SHOULD NOT be merged. In fact, it should be moved the other direction - so that accessibility can be made nil on queries. It is simply not required for reads in most cases.

koji-1009:
https://github.com/mogol/flutter_secure_storage/issues/711#issuecomment-2184934919

About the code reading I did regarding this PR.

dballance:
Yes, the issue is not on writes.

Writes have had accessibility set for several years. It's the addition of `kSecAttrAccessible` on all the other calls that is causing an issue. 

Previously, values were read with a `nil` value for `kSecAttrAccessible`. Now, it's not possible to read with `nil` - you MUST provide a value.

In some cases (many, I'd argue), you can read / delete without setting `kSecAttrAccessible` that was used to store the value. 

I want to be able to READ and DELETE without regard to the `kSecAttrAccessible` value set for this parameter. I COULD do that before, I CANNOT do that now, and I certainly cannot do it if this PR is merged.

koji-1009:
"PlatformException(Unexpected security result code, Code: -25299, Message: The specified item already exists in the keychain. This is a problem that occurs during `write`. This problem appears to be an error at `write` time; the `read` and `delete` cases are a different problem, no?

You may be interested in the discussion of #719 where accessibility was set to `nil` for `read` and `delete`. #719 addresses the issue in #692. We must address multiple issues simultaneously.
This PR corresponds to the code modified in #719. Therefore, once #719 is reverted, this PR will no longer be needed.

dballance:
Yeah, it's not likely to be reverted - #751 should be pulled in which will make this PR moot.

juliansteenbakker:
Closing this in favor of #751

PR #736, opened 2024-06-11
Add useDataProtectionKeyChain parameter to macOS options

Skycoder42:
This pull request aims to enable the optional use of the "classic" macOS keychain. This is done by adding the `useDataProtectionKeyChain` parameter to the macOS options. It defaults to `true` so the default behavior does not change.

The reason for adding this is that the data protection keychain while being more secure has two drawbacks:
1. The entries cannot be accessed with the `security` command line tool
2. The data protection key chain requires an entitlement, which in turn requires the app to be signed. So using this plugin for an unsigned, internal macOS app is not possible without this fix.

I have created 2 commits. The first implements the change described above. The second commit makes the `synchronizable` and `useDataProtectionKeyChain` parameters inside the macOS implementation non-nullable, as they are always set by the code. This simplifies the code as the unnecessary null checks can be removed. If you prefer to keep the nullable booleans or would like two separate pull request, and can revert the second commit.

### DIFF
--- a/flutter_secure_storage/lib/options/macos_options.dart
+++ b/flutter_secure_storage/lib/options/macos_options.dart
@@ -7,7 +7,9 @@ class MacOsOptions extends AppleOptions {
     String? accountName = AppleOptions.defaultAccountName,
     KeychainAccessibility accessibility = KeychainAccessibility.unlocked,
     bool synchronizable = false,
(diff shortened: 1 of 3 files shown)


juliansteenbakker:
This looks good, thanks!

PR #737, opened 2024-06-12
Check contains key with readed value

koji-1009:
The error #711 indicates that `SecItemAdd` is being called when it should be `SecItemUpdate` in some users. This could be a case where (for some reason) `errSecItemNotFound` is returned from `SecItemCopyMatching`, but the value is actually saved.

Therefore, instead of `SecItemCopyMatching`, it has been modified to determine if the key is contains based on the value obtained by `read`.
This implementation is based on v9.0.0.

https://github.com/mogol/flutter_secure_storage/compare/v9.0.0...v9.1.0#diff-9ddc98846efcc4fe8d873eb548ed640010457abc988d46a2d2e2d421f8d29045L38

---

I would like to get your opinion on the modification.

### DIFF
--- a/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
+++ b/flutter_secure_storage/ios/Classes/FlutterSecureStorage.swift
@@ -57,20 +57,6 @@ class FlutterSecureStorage{
         return keychainQuery
     }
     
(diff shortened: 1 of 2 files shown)


rvdveen:
@juliansteenbakker could you retry the failing action? It succeeds on my machine, would be good to have this fix in this awesome library!

koji-1009:
Closed as fixed in #751.

PR #751, opened 2024-07-09
Fix for issue #711: The specified item already exists in the keychain, iOS Code -25299, -25300

frankteller-de:
My suggestion for fixing bugs 25299 and 25300, which have existed on iOS since version greater than 9.0.0.

Original issue: The specified item already exists in the keychain, IOS #711

The problem occurred when the parameter kSecAttrAccessible was used to check whether a certain key already exists. The parameters kSecAttrAccount and kSecAttrAccessible together are not unique. The key kSecAttrAccount can therefore only occur once, even if kSecAttrAccessible is different.

In addition, kSecAttrAccessible cannot be changed by SecItemUpdate. This requires the key to be deleted and recreated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 A Flutter plugin to store data in secure storage:
 
 - [Keychain](https://developer.apple.com/library/content/documentation/Security/Conceptual/keychainServConcepts/01introduction/introduction.html#//apple_ref/doc/uid/TP30000897-CH203-TP1) is used for iOS
-- AES encryption is used for Android. AES secret key is encrypted with RSA and RSA key is stored in [KeyStore](https://developer.android.com/training/articles/keystore.html).   
-  By default following algorithms are used for AES and secret key encryption: AES/CBC/PKCS7Padding and RSA/ECB/PKCS1Padding  
-  From Android 6 you can use newer, recommended algoritms:  
-  AES/GCM/NoPadding and RSA/ECB/OAEPWithSHA-256AndMGF1Padding  
+- AES encryption is used for Android. AES secret key is encrypted with RSA and RSA key is stored in [KeyStore](https://developer.android.com/training/articles/keystore.html).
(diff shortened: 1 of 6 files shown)


dballance:
This is the correct PR to merge. Please do NOT merge #734 as it will exacerbate the issue. 

The parameters that were optional (`kSecAttrAccessible`) should be made optional again. This PR does that, and should be merged to fix the issues in #711

paulking86 (approved):
Excellent work @frankteller-de 👏

dballance:
I think this PR may also be missing the ability to set the accessibility key to `null` via dart. Otherwise, the value is still always set.

frankteller-de:
> I think this PR may also be missing the ability to set the accessibility key to `null` via dart. Otherwise, the value is still always set.

Hi, why do you think that? In this PR, accessibility should be optional.

```swift
if (accessibility != nil) {
    keychainQuery[kSecAttrAccessible] = parseAccessibleAttr(accessibility: accessibility)
}
```

If it is not specified in flutter, the attribute is ignored. 
But at the latest apple will set kSecAttrAccessibleWhenUnlocked as the default attribute. 

Or have I missed something?

juliansteenbakker:
Hi all, if i read everything correctly, this PR is done for merge, however the only thing on the dart side `AppleOptions` that should change is this:
```dart
    KeychainAccessibility? accessibility,
    // KeychainAccessibility accessibility = KeychainAccessibility.unlocked,
```

I will make this change in a new major version since it would change the current default behavior.

paulking86:
@dballance Has kindly done some work on this already @juliansteenbakker. This is the commit you want I believe!

https://github.com/Tango-Tango/flutter_secure_storage/commit/400f6b4eb6ebaea6936fdbaa0d2f16ef36576678

juliansteenbakker:
Thank you for pointing me in the good direction @paulking86 ! Since that change is not breaking, i will merge it right away.

PR #780, opened 2024-08-28
Migrate to win32 5.5.4

### LINKED ISSUE #779: [flutter_secure_storage_windows] Migrate to `win32 5.5.1` or higher
When using `flutter_secure_storage 9.2.2` (which depends on `flutter_secure_storage_windows 3.1.2`),
the version of `flutter_secure_storage_windows` requires a version of win32 that is not compatible with Dart 3.4.0 / Flutter 3.22.0
(and in effect also the latest stable, 3.5.0 / 3.24.0)

We should upgrade `flutter_secure_storage_windows` to that new win32 version.

The reason for this is that we get the following error, due to a change in Dart 3.4.0

```
../../../.pub-cache/hosted/pub.dev/win32-5.0.9/lib/src/guid.dart:32:9: Error: Type 'UnmodifiableUint8ListView' not found.
  final UnmodifiableUint8ListView bytes;
        ^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/win32-5.0.9/lib/src/guid.dart:32:9: Error: 'UnmodifiableUint8ListView' isn't a type.
  final UnmodifiableUint8ListView bytes;
        ^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/win32-5.0.9/lib/src/guid.dart:52:17: Error: Method not found: 'UnmodifiableUint8ListView'.
    return Guid(UnmodifiableUint8ListView(guid));
                ^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/win32-5.0.9/lib/src/guid.dart:56:31: Error: Method not found: 'UnmodifiableUint8ListView'.
  factory Guid.zero() => Guid(UnmodifiableUint8ListView(Uint8List(16)));
                              ^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/win32-5.0.9/lib/src/guid.dart:100:17: Error: Method not found: 'UnmodifiableUint8ListView'.
    return Guid(UnmodifiableUint8ListView(Uint8List.fromList(guidAsBytes)));
                ^^^^^^^^^^^^^^^^^^^^^^^^^
```

navaronbracke:
This PR migrates `flutter_secure_storage_windows` to support Dart 3.4.0 / Flutter 3.22.0 (and by extension also Dart 3.5.0 / Flutter 3.24.0)

I did not update the version of `flutter_secure_storage_windows`, as changing the supported Flutter version could possibly be seen as a breaking change. That said, if we use the new version constraint for the latest Flutter stable across all of `flutter_secure_storage`, you could also start cleaning up the deprecated usages of `describeEnum()` (which are the only ignores in this plugin's Dart code AFAIK)

Fixes #779

### DIFF
--- a/flutter_secure_storage_windows/example/pubspec.yaml
+++ b/flutter_secure_storage_windows/example/pubspec.yaml
@@ -3,8 +3,8 @@ description: Demonstrates how to use the flutter_secure_storage_windows plugin.
 publish_to: 'none'
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.0.0"
+  sdk: '>=3.4.0 <4.0.0'
+  flutter: '>=3.22.0'
 
 dependencies:
   flutter:
--- a/flutter_secure_storage_windows/lib/src/flutter_secure_storage_windows_ffi.dart
+++ b/flutter_secure_storage_windows/lib/src/flutter_secure_storage_windows_ffi.dart
@@ -246,9 +246,7 @@ class DpapiJsonFileMapStorage extends MapStorage {
 
         if (plainTextBlob.ref.pbData.address == NULL) {
           throw WindowsException(
-            // TODO: New member requires win32 ^5.4.0
-            // ignore: deprecated_member_use
-            ERROR_OUTOFMEMORY,
+            WIN32_ERROR.ERROR_OUTOFMEMORY,
             message: 'Failure on CryptUnprotectData()',
           );
         }
@@ -350,9 +348,7 @@ class DpapiJsonFileMapStorage extends MapStorage {
 
       if (encryptedTextBlob.ref.pbData.address == NULL) {
         throw WindowsException(
-          // TODO: New member requires win32 ^5.4.0
-          // ignore: deprecated_member_use
-          ERROR_OUTOFMEMORY,
+          WIN32_ERROR.ERROR_OUTOFMEMORY,
           message: 'Failure on CryptProtectData()',
         );
       }
--- a/flutter_secure_storage_windows/pubspec.yaml
+++ b/flutter_secure_storage_windows/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   flutter_secure_storage_platform_interface: ^1.1.0
   path: ^1.8.0
   path_provider: ^2.0.0
-  win32: ^5.0.0
+  win32: ^5.5.4
 
 dev_dependencies:
   flutter_test:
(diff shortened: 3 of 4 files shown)


flutter_secure_storage_windows/example/pubspec.yaml
```diff
@@ -3,8 +3,8 @@ description: Demonstrates how to use the flutter_secure_storage_windows plugin.
 publish_to: 'none'
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.0.0"
+  sdk: '>=3.4.0 <4.0.0'
```
navaronbracke, line 6: This aligns the windows example with the windows implementation.

flutter_secure_storage_windows/lib/src/flutter_secure_storage_windows_ffi.dart
```diff
@@ -246,9 +246,7 @@ class DpapiJsonFileMapStorage extends MapStorage {
 
         if (plainTextBlob.ref.pbData.address == NULL) {
           throw WindowsException(
-            // TODO: New member requires win32 ^5.4.0
-            // ignore: deprecated_member_use
-            ERROR_OUTOFMEMORY,
+            WIN32_ERROR.ERROR_OUTOFMEMORY,
```
navaronbracke, line 249: These TODO's can now be addressed.

flutter_secure_storage_windows/pubspec.yaml
```diff
@@ -4,8 +4,8 @@ repository: https://github.com/mogol/flutter_secure_storage
 version: 3.1.2
 
 environment:
-  sdk: '>=2.12.0 <4.0.0'
-  flutter: ">=2.0.0"
+  sdk: '>=3.4.0 <4.0.0'
```
navaronbracke, line 7: Since the last two Flutter releases (3.22.0 and 3.24.0) are way ahead of the current version range, I opted into updating it here, to match the Dart 3.4.0 requirement for win32.

chrisdedev:
I feel like this should be merged since it breaks other platform builds. This issue is effecting my applications that I develop for platforms that don't include windows. Since flutter_secure_storage now contains all supported platforms in one pub this affects the build process on any platform.

juliansteenbakker:
Hi, thank you for your work! I think its about time we're migrating to a more recent version of the win32 plugin together with a higher minimum dart SDK requirement.

Issue #761, opened 2024-07-29
Null check operator used on a null value


YawarOsman:
I have used the beta version of the package `flutter_secure_storage_web: ^2.0.0-beta.1` and it works fine on `localhost` but when i host it on a locale network and try to access it through the ip address it shows this error:

> Null check operator used on a null value

I just used write method to save the value:

```
  Future<void> setToken(String value) async {
    const storage = FlutterSecureStorage();
    await storage.write(key: 'token', value: value);
  }
  ```

seealex:
same to me.

Ri-car-do-1989:
Got the same issue. It works OK on Chrome and Firefox in Windows PC, but I got the error while using Chrome in an Android phone. In the same phone, when using Firefox, this issue does not happen but it also doesn't save anything.

Ri-car-do-1989:
Upon further research, I found @[this](https://github.com/juliansteenbakker/flutter_secure_storage/issues/320#issuecomment-976308930). That was my problem, and running my app with HTTPS did solve the issue. I noticed that this information is present in the github page, but it is not present in the pub.dev package page. I'd suggest adding that info in the pub.dev page, as I think many people will look there for information (as I did).

PR #782, opened 2024-08-31
Fix null check error and improve error handling in web implementation

BilalRehman08:
This PR addresses the issue #761 where a "Null check operator used on a null value" error occurs when accessing the storage through an IP address on a local network.

Changes made:
- Modified the `read` method in `FlutterSecureStorageWeb` to handle potential null values more safely
- Added null check before attempting to decrypt values
- Implemented error handling for the decryption process


juliansteenbakker:
Hi, thank you for your work!

It seems as if the null check is not neccesary, because this is already checked in the decrypt method. I have moved the try catch to the decrypt method, so it applies to all function calls. I have created a new PR #847 for this. so i'm closing this one.

PR #791, opened 2024-09-20
[iOS] Use serial queue for execution of iOS keychain operations

adamgic:
**Reason:**
For iOS Keychain write function checks for existence of an item before performing changes, then it either:
- adds new item (if item for key does not exist)
- updates item if it exists

This makes write not an atomic operation on the storage. For performing operation we are using `DispatchQueue.global(qos: .userInitiated)`, which returns concurrent queue. In case two concurrent writings happening on the same key, one of these may fail - checks for existence of entry for key, entry does not exist yet, tries to add new entry, but in the meantime concurrent write stores the entry and the add fails.

**Change:**
This PR changes DispatchQueue used for executing iOS Keychain operations: Use serial queue instead of concurrent for execution of iOS keychain operations to avoid concurrent writing issues.

**Background:**
Before this plugin was using main DispatchQueue, which was serial as well, but it is suggested by Apple documentation not to use main Thread to avoid blocking of the UI. For details see this PR: #536


diogosequeira94:
lgtm

juliansteenbakker:
Thank you for your work! LGTM!

Issue #795, opened 2024-09-24
[DOCS][WINDOWS] Difference between pub.dev documentation and project README


Johan-Carinus:
## Issue
When looking at the README file for the 9.2.1 release tag, it is shown that Windows supports both readAll and deleteAll functions.

<img src="https://github.com/user-attachments/assets/b68d70b2-f91d-40d5-9c7a-ff7856b66d55" width="600">

[https://github.com/mogol/flutter_secure_storage/tree/v9.2.1?tab=readme-ov-file#platform-implementation](https://github.com/mogol/flutter_secure_storage/tree/v9.2.1?tab=readme-ov-file#platform-implementation)

This is in contrast with the documentation that is listed for the pub.dev release:

<img src="https://github.com/user-attachments/assets/e7647c55-a4bb-4933-8957-36d87f98c37a" width="600">

[https://github.com/mogol/flutter_secure_storage/tree/v9.2.1/flutter_secure_storage#configure-windows-version](https://github.com/mogol/flutter_secure_storage/tree/v9.2.1/flutter_secure_storage#configure-windows-version)

I did a quick check in the Windows release notes and it seems like the readAll and writeAll functions were introduced as far back as flutter_secure_storage_windows 1.1.0

## Expected behaviour

The changes in regards to the readAll and writeAll functions should have already been documented in the pub.dev documentation.

## Suggested fix

I've opened this PR (#796) to fix the issue.

Johan-Carinus:
Since the documentation has been updated, this is no longer valid.

PR #796, opened 2024-09-24
[DOCS][WINDOWS] Difference between pub.dev documentation and project README

Johan-Carinus:
As listed in this issue #795, I've updated the documentation used for the pub.dev documentation using what is in the project root README.md file,

Sorry for the very small and fussy change, but I'm hoping that every small bit helps 🙂

### DIFF
--- a/flutter_secure_storage/README.md
+++ b/flutter_secure_storage/README.md
@@ -115,7 +115,7 @@ parts:
 
 ## Configure Windows Version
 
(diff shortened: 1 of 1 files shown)


juliansteenbakker:
Hi, thank you for your change, but i have updated all readme's, so this is no longer valid.

Johan-Carinus:
@juliansteenbakker Thank you for your updates 🙂

Issue #769, opened 2024-08-13
[Android] [Beta] Version 10 - Full migration to custom implementation of SecureSharedPreferences


juliansteenbakker:
A long outstanding problem with the Android part of this package is the deprecated usage of older cryptography usage (CBC with PKCS5/PKCS7 padding) for SDK < 23, and the already deprecated JetSec Crypto library that is being used for the current implementation of SecureSharedPreferences.

To fix these issues, i am moving to a custom implementation of the JetSec Crypto library, and removing all other deprecated methods of encrypting the shared preferences. This requires the minimum SDK to be raised from 21 to 23.

I am currently testing the new implementation in the [branch `version-10`](https://github.com/mogol/flutter_secure_storage/tree/version-10). Any help is greatly appreciated.

Affected issues:
#512 #522 #562 #584 #627 #672 #694 #713 #729 #758 #759

paulking86:
Hi @juliansteenbakker - Thanks for your work on this!

Have just tested the `version-10` branch using the override for `flutter_secure_storage_web: ^2.0.0-beta.1`.

When running the app on a device which has written to secure storage prior to this upgrade, my app crashes on startup (when I call `readAll()`). Here is the traceback:

```
PlatformException(Exception encountered, readAll, java.lang.SecurityException: Could not decrypt key. decryption failed
	at com.it_nomads.fluttersecurestorage.crypto.EncryptedSharedPreferences.decryptKey(EncryptedSharedPreferences.java:566)
	at com.it_nomads.fluttersecurestorage.crypto.EncryptedSharedPreferences.getAll(EncryptedSharedPreferences.java:377)
	at com.it_nomads.fluttersecurestorage.FlutterSecureStorage.readAll(FlutterSecureStorage.java:54)
	at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin$MethodRunner.run(FlutterSecureStoragePlugin.java:156)
	at android.os.Handler.handleCallback(Handler.java:959)
	at android.os.Handler.dispatchMessage(Handler.java:100)
	at android.os.Looper.loopOnce(Looper.java:232)
	at android.os.Looper.loop(Looper.java:317)
	at android.os.HandlerThread.run(HandlerThread.java:85)
Caused by: java.security.GeneralSecurityException: decryption failed
[...]

PR #814, opened 2024-11-04
Set minSdkVersion 23 and delete old encryption codes

koji-1009:
ref #769

---

Two updates are planned in flutter_secure_storage v10.

1. update min sdk to 23 and remove old encryption codes
2. remove EncryptedSharedPreferences that have been deprecated

This PR is a cutout of only the part about updating the min sdk to 23.

Many of the problems associated with #769 are related to old encryption codes. On the other hand, the min sdk has been updated to 21 since flutter 3.22, so there is less reason to support older Android OS.

https://docs.flutter.dev/release/breaking-changes/android-kitkat-deprecation

For this reason, can we proceed with updating the min sdk and replacing EncryptedSharedPreferences separately?
If we can update the min sdk to 23 and release it as flutter_secure_storage v9.3.0, we can close the issue about the old encryption codes.


koji-1009:
In order to support this, the KeyCipher and StorageCipher settings must be set to the new ones in advance. Since the developer needs to update the options in advance, I assume that the PR cannot be merged. For this reason, I have closed the PR.

PR #822, opened 2024-11-11
feat: Update default algorithms

koji-1009:
In preparation for raising the supported version to Android 23, I suggest updating the default values for the options.


juliansteenbakker:
Thank you for your work!

Because version 10 will be the last version that uses this method of encryption, i am not merging this in order to keep it backwards compatible with people that want to migrate their data. Version 11 will only have safe methods of encryption.

PR #824, opened 2024-11-13
feat: Support sessionStorage option for web

rjs580:
Added the ability to use `sessionStorage` instead of `localStorage` through a new option in the WebOptions class. This change involves updating the storage access methods to choose between `sessionStorage` and `localStorage` based on the provided options.

Justification for this change:

1. Scope Control
    - `sessionStorage` is designed to store data for the duration of a page session. This means that when the user closes the browser tab, the data is automatically cleared, which can be advantageous for sensitive information that should not persist after the session ends.

2. Security
    - Storing sensitive information such as authentication tokens in `sessionStorage` can enhance security. It minimizes the risk of leaking sensitive data if another user gains access to the device because the data does not persist between sessions.

3. Use Case Flexibility
    - By providing the option to choose between `localStorage` and `sessionStorage`, developers can tailor data storage strategies based on the specific needs of their application. Long-term storage can use `localStorage`, whereas temporary, session-based data can use `sessionStorage`.

4. Compliance Requirements
    - Some applications need to comply with strict data storage regulations, which may mandate that sensitive information not persist beyond a session. `sessionStorage` supports adherence to such compliance requirements.

### DIFF
--- a/flutter_secure_storage/lib/options/web_options.dart
+++ b/flutter_secure_storage/lib/options/web_options.dart
@@ -7,6 +7,7 @@ class WebOptions extends Options {
     this.publicKey = 'FlutterSecureStorage',
     this.wrapKey = '',
     this.wrapKeyIv = '',
+    this.useSessionStorage = false,
   });
 
   static const WebOptions defaultOptions = WebOptions();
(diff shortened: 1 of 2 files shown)


rjs580:
Hello @juliansteenbakker,

Could you please review and approve this PR? The changes are crucial for our next update, as they introduce important enhancements to data storage options that support improved security and compliance. Your prompt review would be greatly appreciated.

Thank you

juliansteenbakker:
Hi, thank you for your work and sorry for my delay. I am reviewing it now and will include it in a release later today.

Issue #701, opened 2024-04-30
Could not find gradle-8.1.1.jar (com.android.tools.build:gradle:8.1.1)


mustafa-fahimi:
Could not find gradle-8.1.1.jar (com.android.tools.build:gradle:8.1.1)

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

mustafa-fahimi:
Still exists

juliansteenbakker:
Do you have any more information about how to reproduce this?

juliansteenbakker:
This problem is probably fixed with version v10.0.0-beta.1. If you still encounter it with this version, please let me know.

Issue #843, opened 2024-12-27
Strange dependency conflict


malibu1966:
Hi there, I'd like to know how the dependencies changed between v4.2.1 and v5.0.0.  For some reason, there are a couple packages that I use that, once installed, will cause flutter pub add to install only the version 4.2.1 of flutter_secure_storage.  Before I install the packages 'connectanum' and 'build_runner', flutter pub selects v9.2.2 of flutter_secure_storage, but after I install those, only 4.2.1 is selected.  I can also remove these packages and then install flutter_secure_storage at 9.2.2 again.

Any thoughts on why this may be happening?  I'm going to install the other packages after so I can use 9.2.2 and hope that works, but I would like to know why this behavior may be happening.

Thank you.

NeoFahrenheit:
I got a similar problem:
https://stackoverflow.com/questions/79313061/cant-install-recent-version-of-flutter-secure-storage-on-flutter

juliansteenbakker:
This problem is probably fixed with version v10.0.0-beta.1. If you still encounter it with this version, please let me know.

Issue #839, opened 2024-12-18
Adding `flutter_secure_storage` to a new project causes build failures with: ` Missing classes detected while running R8. `


arran4:
With a brand new project:
```
arran@arran-desktop:/home/arran/Documents/Projects/untitled2 34340 
% java --version
openjdk 17.0.13 2024-10-15
OpenJDK Runtime Environment 17.0.13_p11 (build 17.0.13+11)
OpenJDK 64-Bit Server VM 17.0.13_p11 (build 17.0.13+11, mixed mode, sharing)
arran@arran-desktop:/home/arran/Documents/Projects/untitled2 34341 
% flutter --version
Flutter 3.27.1 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 17025dd882 (2 days ago) • 2024-12-17 03:23:09 +0900
Engine • revision cb4b5fff73
Tools • Dart 3.6.0 • DevTools 2.40.2
```

```
arran@arran-desktop:/home/arran/Documents/Projects/untitled2 34369 
% grep -v '^\(\s*#\|$\)' pubspec.yaml
name: untitled2
description: "A new Flutter project."
publish_to: 'none' # Remove this line if you wish to publish to pub.dev
version: 1.0.0+1
environment:
  sdk: ^3.6.0
dependencies:
  flutter:
    sdk: flutter
  flutter_secure_storage: 9.2.2
dev_dependencies:
  flutter_test:
    sdk: flutter
  flutter_lints: ^5.0.0
flutter:
  uses-material-design: true
```

Causes:
```
arran@arran-desktop:/home/arran/Documents/Projects/untitled2 34368 
% flutter clean && flutter build appbundle
Deleting build...                                                   25ms
Deleting .dart_tool...                                               3ms
Deleting Generated.xcconfig...                                       0ms
Deleting flutter_export_environment.sh...                            0ms
Deleting ephemeral...                                                0ms
Deleting ephemeral...                                                0ms
Deleting ephemeral...                                                0ms
Deleting .flutter-plugins-dependencies...                            0ms
Deleting .flutter-plugins...                                         0ms
Resolving dependencies... 
Downloading packages... 
  async 2.11.0 (2.12.0 available)
  boolean_selector 2.1.1 (2.1.2 available)
  characters 1.3.0 (1.4.0 available)
  clock 1.1.1 (1.1.2 available)
  collection 1.19.0 (1.19.1 available)
[...]

Issue #830, opened 2024-12-02
Migrate the Package to support AGP 8+


KhushalJangid:
Flutter version: 3.24.5
Java version: 21
Gradle version: 8.5

The flutter engine throws this error:

A problem occurred configuring project ':flutter_secure_storage'.
> Could not create an instance of type com.android.build.api.variant.impl.LibraryVariantBuilderImpl.
   > Namespace not specified. Specify a namespace in the module's build file. See https://d.android.com/r/tools/upgrade-assistant/set-namespace for information about setting the namespace.

We will need to add this `namespace` property to the build.gradle of the package

juliansteenbakker:
This problem is probably fixed with version v10.0.0-beta.1. If you still encounter it with this version, please let me know.

PR #848, opened 2025-01-06
feat: release of v10.0.0-beta.1

juliansteenbakker:
This new major release has some big changes. This plugin requires a minimum dart sdk of 3.3.0 or higher
and a minimum flutter version of 3.19.0.

[Android]
- By default, encryptedSharedPreferences will be enabled, and cannot be disabled. If there is still 
  data saved by previous versions using encryptedSharedPreferences = false, it will be automatically
  transferred to encryptedSharedPreferences.
- Migrated from deprecated Jetpack Crypto library to Google Tink Crypto library.
- Migrated to Android SDK 35
- Migrated to Java Version 17
- Minimum Android SDK is changed from 19 to 23.
- Lots of minor code improvements

[iOS]
- Use serial queue for execution of iOS keychain operations

[Web]
- Web is now migrated to be compatible with WASM.
- The parameter useSessionStorage is added to WebOptions, which you can use to save in session storage
  instead of local storage.

[Windows]
- Migrates to `win32` version 5.5.4 to support Dart 3.4 / Flutter 3.22.0.


juliansteenbakker:
Fixes #701, #748, #769, #789, #794, #803, #806, #818, #826, #830, #839,  #843

PR #859, opened 2025-01-09
Reintroduce deprecation warning supression.

### LINKED ISSUE #855: v10.0.0-beta.2 - MasterKey.java uses or overrides a deprecated API.
Hi, after updating to `v10.0.0-beta.2` I am unfortunately getting the following compile warning when doing a release build (`flutter run --release`) after `flutter clean`:
```
Note: C:\myPubCache\flutter_secure_storage-10.0.0-beta.2\src\main\java\com\it_nomads\fluttersecurestorage\crypto\MasterKey.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details
```

When ran with `-Xlint:deprecation` I get the following output:
```
C:\myPubCache\flutter_secure_storage-10.0.0-beta.2\src\main\java\com\it_nomads\fluttersecurestorage\crypto\MasterKey.java:298: warning [deprecation] setUserAuthenticationVaildityDurationSeconds(int) in Builder has been deprecated

keyGenBuilder.setUserAuthenticationVaildityDurationSeconds(
```

I suspect, that [this commit](https://github.com/juliansteenbakker/flutter_secure_storage/commit/aa12d959324b9927254eab377cfff30437401fdd#diff-6a46206c00b12ddb5cc124f2e3f37afc8199e372b9f51f49507b58e95f19bdc4) makes the warning appear. 

Is it ok if I provide a PR that reintroduces the `@SuppressWarnings("deprecation")` annotation?

FrankenApps:
Fixes #855 for now.

Maybe in the future we could migrate the code like [explained here](https://stackoverflow.com/a/66067487/4593433). But I would have to look into this more thoroughly, in order to not accidentially increase the minSdkLevel.

### DIFF
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/crypto/MasterKey.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/crypto/MasterKey.java
@@ -276,6 +276,7 @@ public final class MasterKey {
             static String getKeystoreAlias(KeyGenParameterSpec keyGenParameterSpec) {
                 return keyGenParameterSpec.getKeystoreAlias();
             }
+            @SuppressWarnings("deprecation")
             static MasterKey build(Builder builder) throws GeneralSecurityException, IOException {
                 if (builder.mKeyScheme == null && builder.mKeyGenParameterSpec == null) {
                     throw new IllegalArgumentException("build() called before "


juliansteenbakker:
Thank you for the PR!

All deprecated functions are already checked. The function that is causing this warning is https://github.com/juliansteenbakker/flutter_secure_storage/blob/d68dd9bc7f6fc8f27ab58237cfa851009643a619/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/crypto/MasterKey.java#L294, and as you can see it already has a check for it. I don't know why the compiler keeps bragging about it, because Android Studio can't detect any issues with it.

FrankenApps:
Oh, I see now. Yeah that is strange.

PR #886, opened 2025-02-10
docs(Linux): remove mention of no longer needed libjsoncpp from README

cyberbobs:
Remove reference to no longer needed libjsoncpp on Linux from README.


juliansteenbakker:
Thanks, LGTM!

PR #906, opened 2025-03-08
[Android] Optimize Secure Storage by Replacing Single Worker Thread with Thread Pool

farmery:
### Problem
Secure storage operations were previously executed on a single worker thread. This becomes a bottleneck when multiple read operations are triggered simultaneously, leading to significant delays due to the synchronous execution on the native side.

### Solution
Replaced the single worker thread with a thread pool (using Executors.newCachedThreadPool()). This allows multiple secure storage operations to run concurrently, significantly improving performance when handling simultaneous read/write requests.

### Impact
Reduces contention and latency for secure storage operations.
Improves app responsiveness when multiple storage operations are triggered.
No changes in API usage—just an internal optimization for Android.

### DIFF
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
@@ -1,7 +1,6 @@
 package com.it_nomads.fluttersecurestorage;
 
 import android.os.Handler;
-import android.os.HandlerThread;
 import android.os.Looper;
 
 import androidx.annotation.NonNull;
@@ -10,6 +9,8 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import io.flutter.embedding.engine.plugins.FlutterPlugin;
 import io.flutter.plugin.common.MethodCall;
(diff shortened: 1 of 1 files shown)


farmery:
HI @juliansteenbakker, please can you take a look?🙏🏾

farmery:
@juliansteenbakker Hii...👀

flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -21,30 +22,30 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
...
 
     @Override
     public void onAttachedToEngine(FlutterPluginBinding binding) {
         this.binding = binding;
-        workerThread = new HandlerThread("fluttersecurestorage.worker");
-        workerThread.start();
-        workerThreadHandler = new Handler(workerThread.getLooper());
+        int availableCores = Runtime.getRuntime().availableProcessors();
+        int threadCount = Math.min(4, Math.max(2, availableCores));
+        this.executor = Executors.newFixedThreadPool(threadCount);
```
juliansteenbakker, line 33: Why use a fixed thread, when we can also use newCachedThreadPool which in my opinion seems more useful for these i/o actions? This also removes the need for looking up cores and threads.

juliansteenbakker (commented):
Sorry for the delay, see my comment on changelog of v10.0.0-beta.5 for some explanation. I have one question about this implementation.

juliansteenbakker:
I have given it some more thought, but i don't think we should switch to a threaded system, because this can introduce race conditions where you can perform multiple actions for a key and you don't know which one will finish first. It would be better to use readAll, and loop over the results yourself, which you can handle in the background.

Ofcourse if you think otherwise please let me know.

PR #921, opened 2025-04-19
Fix macOS Privacy Manifest Integration to Prevent CFBundleExecutable Validation Error

venkatareddyboya-highspot:
📝 Description:
This PR resolves a macOS App Store Connect validation issue caused by improper usage of s.resource_bundles in the podspec, which led to the error:

Bad CFBundleExecutable. Cannot find executable file that matches the value of CFBundleExecutable...

🔧 Changes Made:
Replaced s.resource_bundles with s.resources in flutter_secure_storage_darwin.podspec to include PrivacyInfo.xcprivacy directly into the framework bundle instead of a nested .bundle.

This ensures that the privacy manifest file is placed correctly without creating a nested resource bundle, which is unnecessary and triggers validation errors during App Store/TestFlight submission.

📌 Why This Fix:
Starting from Xcode 15 and iOS 17/macOS 14, Apple requires a valid [Privacy Manifest (PrivacyInfo.xcprivacy)](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) in frameworks.

Using s.resource_bundles creates a separate bundle (flutter_secure_storage.bundle) without an executable, leading to CFBundleExecutable issues.

Switching to s.resources embeds the privacy file directly inside the framework's Resources folder, complying with App Store expectations.

🧪 Validation:
App submission now passes without CFBundleExecutable error.

Verified privacy manifest appears correctly under the framework's Resources.

Let me know if you'd like me to reword or add anything else!

### DIFF
--- a/flutter_secure_storage/pubspec.yaml
+++ b/flutter_secure_storage/pubspec.yaml
@@ -31,7 +31,8 @@ dependencies:
   # implementation constraints as "any". We cannot do it right now as it fails pub publish
   # validation, so we set a ^ constraint.
   # https://github.com/flutter/flutter/issues/46264
-  flutter_secure_storage_darwin: ^0.1.0
+  flutter_secure_storage_darwin: #^0.1.0
+    path: '../flutter_secure_storage_darwin'
   flutter_secure_storage_linux: ^2.0.0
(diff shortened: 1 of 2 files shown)


juliansteenbakker:
The changes are not valid.

PR #928, opened 2025-05-11
check if window has secureContext

imsamgarg:
I added a condition to check if the window has a secure context. Without this check, debugging the issue will take some time.


juliansteenbakker:
Looks good to me, thanks!

PR #962, opened 2025-07-11
Refactor Android storage cipher implementations

KristijanMitrik:
Removed StorageCipher18Implementation and refactored StorageCipherGCMImplementation to be self-contained, using GCM mode as the default. Updated StorageCipherFactory to simplify algorithm selection, improve enum handling, and remove legacy CBC support. Also updated .gitignore to exclude C++ build artifacts from the example app.


juliansteenbakker:
Thank you for your work. However, i have already rewritten the android part to remove legacy code and migrate to a whole new code base.

PR #975, opened 2025-08-21
Fix: Handle Keystore exceptions on iMin devices

ykhandelwaly:
This change addresses a `GeneralSecurityException: Keystore operation failed` error that occurs on certain devices, notably the iMin POS running Android 10.

The root cause appears to be a device-specific bug in the Android KeyStore implementation that prevents the generation of a standard 256-bit AES key.

To resolve this, a fallback mechanism has been implemented. If the initial attempt to create a 256-bit master key fails, the code will now automatically retry the operation with a 128-bit key. This provides a workaround for the affected devices without altering the behavior for devices that correctly support 256-bit keys.


juliansteenbakker:
Thank you for your fix. However, the part you wrote the fix for will be deprecated in the latest beta version. I think we can still land the fix in the new custom cipher, which generates the key. I will close this PR, and am open to any new one using the custom cipher factory.

PR #979, opened 2025-08-27
Explain the underscore being added automatically to the prefix

lukaszdebowski:
I've spent around 3hrs wondering why I cannot read the value from the storage on the Kotlin side, only to discover that the prefix is actually not a complete prefix, there is also an underscore. This should be noted in the docs.


marcsanny (approved):

juliansteenbakker:
Thanks for the doc update!

Issue #917, opened 2025-04-03
Saved data on iOS is not secure


zigapovhe:
Recently, one of the apps I've been working on underwent a penetration testing review. The pentest discovered (with demonstrated proof) that local storage data on iOS devices can be accessed in plaintext. Since our app uses flutter_secure_storage, this finding was surprising to me.

Initially, I suspected that we might have misconfigured the plugin. Upon reviewing our implementation, I confirmed that we are using the most restrictive Keychain option provided by the plugin:

```dart
final _storage = const FlutterSecureStorage(
      iOptions: IOSOptions(
    accessibility: KeychainAccessibility.unlocked_this_device,
  ));
```
Upon further investigation, I found [this page in Apple's documentation](https://support.apple.com/en-gb/guide/security/secb0694df1a/web), stating that Apple uses two different AES-256-GCM keys for encryption. However, it appears that data encryption occurs only just before writing to disk and is immediately decrypted when accessed.

Thus, while data on the physical disk is indeed encrypted, any jailbroken iOS device can easily retrieve all keychain-stored values in plaintext once the device is unlocked. Notably, the penetration testers neither had access to our source code nor knew the keys used for storing data in the keychain; yet, they were still able to show us all plaintext data.

Their recommendation was to enhance security by generating an asymmetric key stored in the Secure Enclave and using it alongside Apple's built-in encryption mechanisms to achieve true security for keychain-stored data.

Given that flutter_secure_storage already utilizes a similar approach for Android, I'm curious if a similar solution could be implemented for iOS as well.

Thanks!

jeffsheets:
@zigapovhe thanks for starting the discussion, I arrived here after also hearing about pen tester concerns.

I think there are another couple of possible options too that we should explore

One note: the most restrictive option is `KeychainAccessibility.passcode`. From reading the details (https://developer.apple.com/documentation/security/restricting-keychain-item-accessibility), the vulnerability might have been that `A device without a passcode is considered to always be unlocked`.  Do you know if the pen test results would have been the same with `passcode` option used? The plugin here does set [kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessiblewhenpasscodesetthisdeviceonly) for the passcode option.

[...]

PR #989, opened 2025-09-18
Add Secure Enclave Support to iOS/macOS

ahmeddhus:
### Description
This PR adds support for storing keychain items using the Secure Enclave on iOS and macOS.
A new `useSecureEnclave` boolean option is introduced in `IOSOptions` and `AppleOptions`.
When enabled, secrets are stored using Secure Enclave-backed keys if available.
If Secure Enclave is unavailable or unwrapping fails, the plugin falls back to standard keychain storage.

### Changes
- Added `useSecureEnclave` to Dart options (`IOSOptions`, `AppleOptions`)
- Passed the flag to native code via method channel
- Updated iOS/macOS native implementations to use Secure Enclave when requested
- Implemented fallback logic for devices without Secure Enclave
- Updated integration tests to cover Secure Enclave and fallback scenarios
- Updated documentation and code comments

### Motivation
Addresses [#917](https://github.com/juliansteenbakker/flutter_secure_storage/issues/917)
Provides stronger security for sensitive data on supported Apple devices.

### Checklist
- [x]  All tests pass
- [x]  Documentation updated
- [x]  Backwards compatible


flutter_secure_storage/example/integration_test/app_test.dart
```diff
@@ -75,6 +78,130 @@ void main() {
         ..verifyRowDoesNotExist(0)
         ..verifyRowDoesNotExist(1);
     });
+
+    testWidgets('Enclave requested on iOS Simulator falls back gracefully',
+        (WidgetTester tester) async {
+      if (!(Platform.isIOS &&
+          Platform.environment.containsKey('SIMULATOR_DEVICE_NAME'))) {
+        return; // Skip when not running on iOS Simulator
+      }
```
SarcasticNoodle: let me know what you think about this
```suggestion
    testWidgets('Enclave requested on iOS Simulator falls back gracefully',
        skip: !(Platform.isIOS && Platform.environment.containsKey('SIMULATOR_DEVICE_NAME')),
        (WidgetTester tester) async {
```

juliansteenbakker:
Thank you for your contribution! I will test this as soon as i get my apple device up and running.

flutter_secure_storage/example/integration_test/app_test.dart
```diff
@@ -78,6 +81,127 @@ void main() {
...
+        key: key,
+        iOptions: const IOSOptions(useSecureEnclave: true),
+      );
+      expect(afterDelete, isNull);
+    });
+
+    testWidgets(
+        'iOS device: baseline (useSecureEnclave=false) write/read/delete',
+        (WidgetTester tester) async {
+      if (!(Platform.isIOS &&
```
juliansteenbakker: Why not use the `skip` parameter like the test above?

flutter_secure_storage/example/integration_test/app_test.dart
```diff
@@ -78,6 +81,127 @@ void main() {
...
+        key: key,
+        iOptions: IOSOptions.defaultOptions,
+      );
+      expect(afterDelete, isNull);
+    });
+
+    testWidgets(
+        'iOS device: useSecureEnclave=true with non-prompting access control (applicationPassword) write/read/delete',
+        (WidgetTester tester) async {
+      if (!(Platform.isIOS &&
```
juliansteenbakker: Why not use the `skip` parameter like the test above?

juliansteenbakker (changes requested):
Overall implementation looks good, just a minor issue

flutter_secure_storage/example/integration_test/app_test.dart
ahmeddhus: Fixed

flutter_secure_storage/example/integration_test/app_test.dart
ahmeddhus: Fixed

juliansteenbakker: Thank you, i will merge as soon as lint issues are fixed and ci passes

ahmeddhus: All green now ✅

zmuranaka:
@juliansteenbakker Are there any changes this needs left? My company is hoping this comes in a release soon 💪

juliansteenbakker:
I've created some code enabling a smooth migration from and to the secure enclave. I will add this asap, and release it as a whole later this week.

juliansteenbakker:
I have one more question right now about the implementation. It seems that on default settings, 'userPresence' is selected for AccessControlFlags. This causes a touchId prompt for every action. I got 7 prompts when opening the example app with 3 keys.

I think we should use [LAContext](https://developer.apple.com/documentation/localauthentication/lacontext) to prevent this. What do you think?

ahmeddhus:
Hi, I accidentally closed my PR while discarding changes in GitHub’s conflict editor.
The branch still exists, but I don’t have permission to reopen the PR.
Can you please reopen it for me?

@juliansteenbakker

juliansteenbakker:
HI, it says that your develop branch is up to date (the same) as my develop branch, so i can't reopen

ahmeddhus:
Can you check now please?

ahmeddhus:
> I have one more question right now about the implementation. It seems that on default settings, 'userPresence' is selected for AccessControlFlags. This causes a touchId prompt for every action. I got 7 prompts when opening the example app with 3 keys.
> 
> I think we should use [LAContext](https://developer.apple.com/documentation/localauthentication/lacontext) to prevent this. What do you think?

Good catch. I’ve updated the iOS/macOS plugin to reuse a single LAContext per Flutter method call and pass it via kSecUseAuthenticationContext (with a 30s touchIDAuthenticationAllowableReuseDuration). This keeps the userPresence default for Secure Enclave but collapses the multiple Touch ID/Face ID prompts down to one per operation. Let me know if you want a different reuse duration or to change the default flags

juliansteenbakker:
Thank you for the changes! I tried the code, it now gives me 1 popup per action, however, readAll always returns zero items.

ahmeddhus:
> Thank you for the changes! I tried the code, it now gives me 1 popup per action, however, readAll always returns zero items.

Thanks for the feedback! I’ve just pushed a fix, could you check it again and let me know if it works now?

westracer:
I tried to use this new feature. Writing the data works (`dataStatus` = 0), but I'm getting an error when trying to read that new entry (`SecKeyCreateDecryptedData` fails):

```
▿ Optional<CFErrorRef>
  - some : Error Domain=com.apple.LocalAuthentication Code=-1009 "ACL operation is not allowed: 'ock'" UserInfo={NSDebugDescription=ACL operation is not allowed: 'ock', NSLocalizedDescription=Operation is not allowed.}
```

Dart code:
```
secretStorage: const FlutterSecureStorage(
  aOptions: AndroidOptions(enforceBiometrics: true),
  iOptions: IOSOptions(
    accessibility: KeychainAccessibility.passcode,
    accessControlFlags: [AccessControlFlag.userPresence],
    useSecureEnclave: true,
  ),
),
...
await secretStorage.read(key: key);
...
await secretStorage.write(
  key: key,
  value: secret,
);

```

The fallback to "standard read" fails as well, obviously, since the stored data is encrypted (it returns nil string).

Tested on iPhone SE (2nd gen), iOS 18.4, with Touch ID.

westracer:
> I tried to use this new feature. Writing the data works (`dataStatus` = 0), but I'm getting an error when trying to read that new entry (`SecKeyCreateDecryptedData` fails):

After some research, I think the secure enclave private key needs `.privateKeyUsage` access control flag, so it can be used for crypto operations. However, passing it from the Dart side to the whole storage breaks it too 
because it creates another issue with the wrapped key (SecItemAdd returns -25293 authorization error).

Another problem is that the `userPresence` parameter is also applied to the SE private key, and it triggers user authentication 2 times on read: for the SE private key, and for the actual encrypted item (and possibly for the wrapped key as well?). I think it should be triggered only one time for the actual item, and not for the SE key.

My suggested changes are:

1. Make the SE private key use some hardcoded access control params including .privateKeyUsage (or maybe make them customizable?)
2. "deleteAll" method should delete the SE private key as well (or create another method to delete it?)

See [my commit](https://github.com/QuickBirdEng/flutter_secure_storage/commit/3d465546e1dc8c0288ba2a5c410e717a939b1a0b) for my fix implementation

<details>
<summary>Updated ensureEnclavePrivateKey</summary>

```
    private func ensureEnclavePrivateKey(service: String?) throws -> SecKey {
        let tag = enclaveKeyTag(for: service) as CFData

        let query: [CFString: Any] = [
            kSecClass: kSecClassKey,
            kSecAttrKeyType: kSecAttrKeyTypeECSECPrimeRandom,
            kSecAttrApplicationTag: tag,
            kSecReturnRef: true
        ]
        var item: CFTypeRef?
        let status = SecItemCopyMatching(query as CFDictionary, &item)
        if status == errSecSuccess, let item = item {
            return (item as! SecKey)
        }

        var attributes: [CFString: Any] = [
            kSecAttrKeyType: kSecAttrKeyTypeECSECPrimeRandom,
            kSecAttrKeySizeInBits: 256,
            kSecAttrTokenID: kSecAttrTokenIDSecureEnclave,
            kSecPrivateKeyAttrs: [
                kSecAttrIsPermanent: true,
                kSecAttrApplicationTag: tag
            ]
        ]
        // Hardcoded access control
        let accessControl = SecAccessControlCreateWithFlags(
            nil,
            kSecAttrAccessibleWhenUnlockedThisDeviceOnly,
            .privateKeyUsage,
            nil
        )
        if let ac = accessControl {
            var privateAttrs = attributes[kSecPrivateKeyAttrs] as! [CFString: Any]
            privateAttrs[kSecAttrAccessControl] = ac
            attributes[kSecPrivateKeyAttrs] = privateAttrs
        }

        var error: Unmanaged<CFError>?
        guard let privateKey = SecKeyCreateRandomKey(attributes as CFDictionary, &error) else {
            throw OSSecError(status: errSecParam, message: error?.takeRetainedValue().localizedDescription)
        }
        return privateKey
    }
```
</details>

ahmeddhus:
Thanks for the feedback. I've addressed all the issues:
- Fixed SE private key access control: Uses hardcoded .privateKeyUsage instead of user-provided flags, preventing "ACL operation is not allowed" errors.
- Eliminated double authentication: Authentication now only happens once at the data item level.
- Implemented deleteAll cleanup: Added deleteEnclavePrivateKey and integrated it into deleteAll.

This should resolve the decryption errors and double authentication prompts.

@westracer

flutter_secure_storage_darwin/darwin/flutter_secure_storage_darwin/Sources/flutter_secure_storage_darwin/FlutterSecureStorageDarwinPlugin.swift
```diff
@@ -165,9 +167,17 @@ public class FlutterSecureStorageDarwinPlugin: NSObject, FlutterPlugin, FlutterS
             isPlaceholder: (options["isPlaceholder"] as? String).flatMap { Bool($0) },
             shouldReturnPersistentReference: (options["persistentReference"] as? String).flatMap { Bool($0) },
             authenticationUIBehavior: options["authenticationUIBehavior"] as? String,
-            accessControlFlags: options["accessControlFlags"] as? String
+            accessControlFlags: options["accessControlFlags"] as? String,
+            useSecureEnclave: (options["useSecureEnclave"] as? String).flatMap { Bool($0) }
         )
 
+        // Reuse a single authentication context to avoid multiple prompts per call.
+        if #available(iOS 9.0, macOS 10.12, *) {
```
juliansteenbakker: I think its best to wrap this whole block with a check if useSecureEnclave is true, because otherwise it would change behaviour for the current implementation.

flutter_secure_storage_darwin/darwin/flutter_secure_storage_darwin/Sources/flutter_secure_storage_darwin/FlutterSecureStorageDarwinPlugin.swift
```diff
@@ -5,9 +5,11 @@
 #if os(iOS)
 import Flutter
 import UIKit
+import LocalAuthentication
```
juliansteenbakker: We can move this double declaration above or below the if else statement, since we need it for both iOS and macOS.

flutter_secure_storage_darwin/darwin/flutter_secure_storage_darwin/Sources/flutter_secure_storage_darwin/FlutterSecureStorageDarwinPlugin.swift
```diff
@@ -5,9 +5,11 @@
...
 #else
 import AppKit
 import FlutterMacOS
+import LocalAuthentication
```
juliansteenbakker: If the above is moved, this one can be removed.

juliansteenbakker (changes requested):
Thank you again for your hard work! I have two more remarks. If those are solved, i will merge and release it!

ahmeddhus:
Thanks for the review and the detailed suggestions!
 - I’ve updated parseCall so that the LAContext reuse is only enabled when `useSecureEnclave == true`, keeping the behavior unchanged for existing non–Secure Enclave usage.
 - I’ve also moved the LocalAuthentication import above the `#if os(iOS)` / `#else` block and removed the duplicate import, so it’s shared between iOS and macOS.

 Let me know if you’d like me to tweak anything else around the auth context or imports.
 
 @juliansteenbakker

juliansteenbakker (approved):

keithhie:
Dear team, thank you for the great work in bringing Secure Enclave to be part of this wonderful package. May I know when can we expect this to be released as an official version of the package?

connor-goTenna:
Any news on this coming to an official version of?

florianbeaupoilvnv:
Hi Team, I would like to thank you for the amazing work on this repository !

This feature is very interesting, and like others here I was wondering when it might be available in an official release. Do you happen to have a roadmap you could share?

Issue #951, opened 2025-06-25
[Linux] Breaks Gnome Keyring with Empty Password


ruediste:
The secrets are stored as JSON value with multiple keys. The string is pretty formatted, thus has newlines. When using a gnome keyring with an empty password, such a value breaks the loading of the keyring during the next boot. 

Proposed solution: Use Base64 encoding before storing the value.

MrFastDie:
Is there any action on this?

juliansteenbakker:
Thanks for bringing this up, i will dive into this later this week.

juliansteenbakker:
I have released v10 which should fix this issue.

duckzooka:
Independent confirmation of this, with the full mechanism traced end to end. It is worse than a
single value failing to round-trip: **the whole keyring becomes unloadable**, taking every other
application's secrets with it.

### Setup

- Ente Auth (`io.ente.auth`) 4.4.24, which bundles `flutter_secure_storage_linux`
- gnome-keyring `1:50.0-1`, libsecret `0.21.7-1`, Arch Linux (kernel 7.2.3), Omarchy 4.0.3
- Default keyring has an **empty password** — the distro provisions it that way on purpose,
  because it enables display-manager autologin, so PAM never receives a password and
  `pam_gnome_keyring` cannot unlock anything (`gkr-pam: no password is available for user`)

### What happens

`flutter_secure_storage` stores its value as pretty-printed JSON. In my case that is a 194-byte
value containing **4 raw newlines**. gnome-keyring's passwordless keyring format is line-based
INI, with the secret written as a single `secret=` line, so the newlines terminate the record and
the file stops being parseable. Structure as written to disk (values redacted):

```ini
[3]
item-type=0
display-name=io.ente.auth/FlutterSecureStorage
secret={
  "auth_secret_key": "<REDACTED>",
  "key": "<REDACTED>",
  "secret_key": "<REDACTED>",
}
mtime=1788365819
ctime=1788365819
```

The write itself reports success — libsecret is perfectly happy, and the running daemon serves the
value correctly from memory for the rest of the session. The damage only surfaces on the next
daemon start:

```
gnome-keyring-daemon[1060]: keyring was in an invalid or unrecognized format: \
    /home/<user>/.local/share/keyrings/Default_keyring.keyring
```

### Why this is more than one app losing its secret

The daemon rejects the entire file, so every unrelated item in the default collection goes with
[...]

PR #1001, opened 2025-11-15
Use json dump without inserting newlines when saving to keyring on linux

JBiene:
Hi all,

i feel like inserting newlines in the json string in #756 introduced problems.

I was trying to set up Ente Auth (https://github.com/ente-io/ente) which made the Default_keyring unusable in all apps that use the same keyring on each reboot. This issue was also mentioned here #951 and here https://github.com/ente-io/ente/discussions/4769. 

System:
Linux 6.17.7-arch1-2
libsecret 0.21.7-1
gnome-keyring 48.0


graciatheciag:
Good job JBiene, super cool🤩

juliansteenbakker:
I have tested it, and seems good to me! I will add an extra changelog entry, so that users who would like the old behaviour could pin their version.

PR #1031, opened 2026-01-13
[WIP] fix(android): isolate storage instances per shared preferences name (fix: #1023)

### LINKED ISSUE #1023: [Android] Multiple instances with different sharedPreferencesName share cached preferences, causing cross-namespace data operations
## Description

When using multiple `FlutterSecureStorage` instances with different `sharedPreferencesName` configurations on Android, operations like `deleteAll()` may affect data from a different namespace than intended.

## Environment

- flutter_secure_storage: 10.0.0
- Platform: Android

## Steps to Reproduce

```dart
// Instance 1
final storage1 = FlutterSecureStorage(
  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_a'),
);

// Instance 2
final storage2 = FlutterSecureStorage(
  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_b'),
);

// Write data to both
await storage1.write(key: 'token', value: 'value_a');
await storage2.write(key: 'token', value: 'value_b');

// Delete all from storage2
await storage2.deleteAll();

// Expected: storage1 data remains intact
// Actual: storage1 data is also deleted
final result = await storage1.read(key: 'token'); // Returns null
```

## Analysis

In `FlutterSecureStorage.java`, the `initialize()` method caches the `preferences` instance:

```java
protected void initialize(FlutterSecureStorageConfig config, SecurePreferencesCallback<Void> callback) {
    this.config = config;
    if (preferences != null) {
        callback.onSuccess(null);
        return;
    }
    // ... initialization with config.getSharedPreferencesName()
}
```

Once `preferences` is initialized by the first instance, subsequent instances with different `sharedPreferencesName` values will reuse the cached `preferences` object, ignoring their own configuration.

In contrast, the iOS/macOS implementation (`FlutterSecureStorage.swift` and `FlutterSecureStorageDarwinPlugin.swift`) properly passes `KeychainQueryParameters` on every operation:

```swift
private func deleteAll(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
    let (params, _) = parseCall(call)  // Config parsed fresh for each call
    let response = flutterSecureStorageManager.deleteAll(params: params)
    handleResponse(response, result)
}
```

Each iOS operation constructs its query from the provided parameters, ensuring namespace isolation is respected.

## Impact

Using multiple storage instances with different namespaces is a valid use case for organizing sensitive data (e.g., separating user credentials from app-specific tokens). The current Android implementation does not support this pattern reliably.

---

mateusz-pietras:
# Fix Android namespace isolation for `sharedPreferencesName`

Fixes #1023

## Problem

Multiple `FlutterSecureStorage` instances with different `sharedPreferences Name` values shared a single cached `FlutterSecureStorage` instance, causing `deleteAll()` and other operations to affect data across namespaces.

## Solution

Enforce namespace isolation at the plugin layer by routing each `MethodChannel` call to a dedicated `FlutterSecureStorage` instance per `sharedPreferencesName`. This ensures operations on namespace B never affect namespace A.

**Implementation:**

- Plugin maintains a `Map<String, FlutterSecureStorage>` keyed by `sharedPreferencesName`
- `getOrCreateStorage()` retrieves or creates the correct instance per namespace
- All operations (`read`, `write`, `delete`, `deleteAll`, etc.) use the namespace-specific instance

## Testing

✅ Added Android-only integration test that:

- Writes the same key to two namespaces (`namespace_a`, `namespace_b`)
- Calls `deleteAll()` on `namespace_b`
- Asserts the key in `namespace_a` remains intact

## Breaking Changes

None

## Notes

- `SHARED_PREFERENCES_CONFIG_NAME` remains global (stores migration/cipher metadata only). This poses a threat of possible encryption-related issues when two or more storages use different algorithms. 
- Actual stored values are fully isolated per `sharedPreferencesName`
- Matches iOS/macOS behavior where each operation uses namespace-specific parameters

### DIFF
--- a/flutter_secure_storage/README.md
+++ b/flutter_secure_storage/README.md
@@ -168,6 +168,21 @@ For advanced users, all combinations below are supported using the `AndroidOptio
     - `false`: Gracefully degrades if biometrics unavailable
     - `true`: Strictly requires device security (PIN/pattern/biometric), throws exception if unavailable
 
+##### Multiple storage namespaces (Android)
+
+If you need multiple isolated secure storages (for example, per signed-in user), set `sharedPreferencesName` via `AndroidOptions`:
+
+```dart
+final storageA = FlutterSecureStorage(
+  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_a'),
+);
+final storageB = FlutterSecureStorage(
+  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_b'),
+);
+```
+
+On Android, `FlutterSecureStorage` keeps some migration/cipher metadata in a global configuration preferences file (not namespaced). The actual stored values are isolated by `sharedPreferencesName`.
+
 #### Biometric Authentication
 
 Flutter Secure Storage supports biometric authentication (fingerprint, face recognition, etc.) on Android API 23+.
--- a/flutter_secure_storage/README.md
+++ b/flutter_secure_storage/README.md
@@ -168,21 +168,6 @@ For advanced users, all combinations below are supported using the `AndroidOptio
     - `false`: Gracefully degrades if biometrics unavailable
     - `true`: Strictly requires device security (PIN/pattern/biometric), throws exception if unavailable
 
-##### Multiple storage namespaces (Android)
-
-If you need multiple isolated secure storages (for example, per signed-in user), set `sharedPreferencesName` via `AndroidOptions`:
-
-```dart
-final storageA = FlutterSecureStorage(
-  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_a'),
-);
-final storageB = FlutterSecureStorage(
-  aOptions: AndroidOptions(sharedPreferencesName: 'namespace_b'),
-);
-```
-
-On Android, `FlutterSecureStorage` keeps some migration/cipher metadata in a global configuration preferences file (not namespaced). The actual stored values are isolated by `sharedPreferencesName`.
-
 #### Biometric Authentication
 
 Flutter Secure Storage supports biometric authentication (fingerprint, face recognition, etc.) on Android API 23+.
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
@@ -35,6 +35,10 @@ public class FlutterSecureStorage {
 
     private static final String TAG = "FlutterSecureStorage";
     private static final Charset charset = StandardCharsets.UTF_8;
+    // NOTE: This config store is global (not namespaced by sharedPreferencesName).
+    // It holds migration/cipher markers that apply process-wide, even when the app uses
+    // multiple FlutterSecureStorage namespaces via AndroidOptions.sharedPreferencesName.
+    // This is a documented limitation; namespace isolation is handled by the plugin routing.
     private static final String SHARED_PREFERENCES_CONFIG_NAME = "FlutterSecureStorageConfiguration";
 
     private FlutterSecureStorageConfig config;
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorage.java
@@ -32,15 +32,9 @@ import java.util.concurrent.Executors;
 import javax.crypto.Cipher;
 
 public class FlutterSecureStorage {
-
     private static final String TAG = "FlutterSecureStorage";
     private static final Charset charset = StandardCharsets.UTF_8;
-    // NOTE: This config store is global (not namespaced by sharedPreferencesName).
-    // It holds migration/cipher markers that apply process-wide, even when the app uses
-    // multiple FlutterSecureStorage namespaces via AndroidOptions.sharedPreferencesName.
-    // This is a documented limitation; namespace isolation is handled by the plugin routing.
-    private static final String SHARED_PREFERENCES_CONFIG_NAME = "FlutterSecureStorageConfiguration";
-
+    
     private FlutterSecureStorageConfig config;
     @NonNull
     private final Context context;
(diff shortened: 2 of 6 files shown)


mateusz-pietras:
Opened for initial code review as it directly solves the mentioned issue #1023, however, I will be working on introducing the same type of namespace/storage isolation for the `SHARED_PREFERENCES_CONFIG_NAME`. Meanwhile I would appreciate any suggestions regarding the presented code and the config isolation.

mateusz-pietras:
## Important compatibility note (Android)

On some existing installs, upgrading to this fix can make previously-stored values “disappear” **even when the app passes the same `sharedPreferencesName`**.

### Why this can happen

Issue #1023 is caused by the Android plugin reusing a single `FlutterSecureStorage` instance across namespaces and caching the first-initialized `SharedPreferences`. In practice, this could cause values written for namespace **B** to be physically persisted into namespace **A**’s `SharedPreferences` file (or the default `FlutterSecureStorage` prefs), depending on initialization order and app lifecycle.

After this PR, reads/writes are correctly bound to the requested namespace. If legacy data was accidentally stored under a *different* prefs file due to the bug, the upgraded version will not find it in the “correct” namespace anymore (because it never lived there).

In other words: this PR fixes the bug, but it also removes accidental cross-namespace behavior some apps may have relied on unknowingly. Apps that previously experienced namespace mixing might observe missing values after upgrade.

## Suggested solution (introduce a new field; deprecate `sharedPreferencesName`)

To provide a clean, deterministic upgrade path while keeping strong isolation semantics, I suggest introducing a **new Android option** that becomes the canonical “namespace identifier”, and deprecating the existing `sharedPreferencesName` field.

### Proposed API direction

- **New field** (name TBD), e.g. `storageNamespace` / `namespace` / `storageId`
- **Deprecated field**: `sharedPreferencesName`

### Proposed behavior

- The plugin will treat the new field as the source of truth for *all* Android namespacing:
  - The data `SharedPreferences` file name
  - Any config/marker preference stores used for algorithm/migration metadata

- For backwards compatibility during the deprecation window:
  - If the new field is **not** provided, the plugin falls back to the deprecated `sharedPreferencesName` (current behavior).
  - If both are provided, the new field wins (and the old one is ignored, with a log/warn).

### Migration expectations

This does **not** magically recover data that was physically written to the wrong prefs file due to the legacy bug (because that location was non-deterministic). However, introducing the new field provides:
- a clearer contract going forward (“this is the stable namespace identifier”),
- and a controlled breaking-change path where apps can move to the new field and avoid relying on low-level implementation details (`SharedPreferences` file naming).

### Deprecation plan

- Mark `sharedPreferencesName` as deprecated in docs (and API annotations where possible).
- Document the upgrade path: “replace `sharedPreferencesName` with `<newField>`”.
- Keep the fallback behavior for at least one major/minor cycle before removing `sharedPreferencesName`.

westracer:
> SHARED_PREFERENCES_CONFIG_NAME remains global (stores migration/cipher metadata only). This poses a threat of possible encryption-related issues when two or more storages use different algorithms

This is actually a big problem that needs to be solved as well, and not just a "note". It's especially relevant since there's now a migration mechanism in v10 from AES/CBC to another cipher. If you have 2 storages with different algorithms/keys, they will conflict because they are using the same key alias in KeyStore, and all data will be lost as a result. Different keys should be used for different storages

mateusz-pietras:
> > SHARED_PREFERENCES_CONFIG_NAME remains global (stores migration/cipher metadata only). This poses a threat of possible encryption-related issues when two or more storages use different algorithms
> 
> This is actually a big problem that needs to be solved as well, and not just a "note". It's especially relevant since there's now a migration mechanism in v10 from AES/CBC to another cipher. If you have 2 storages with different algorithms/keys, they will conflict because they are using the same key alias in KeyStore, and all data will be lost as a result. Different keys should be used for different storages

Yes, that is true. I introduced a `NamespacedConfigSource` concept in the most recent commit, that allows to use separate algorithms per namespaced storage.  That should solve the issue.

westracer:
> Yes, that is true. I introduced a `NamespacedConfigSource` concept in the most recent commit, that allows to use separate algorithms per namespaced storage. That should solve the issue.

That doesn't solve the issue fully because the key alias is the same for all algorithms/storages, which means the same key is used for all instances:

```
context.getPackageName() + ".FlutterSecureStoragePluginKey"
```

but obviously, keys should be separated as well for different algorithms/storages; otherwise, they will conflict with each other. Correct me if I'm wrong

juliansteenbakker:
Hi all, sorry for the delay. I think a valid issue is raised here. I think the best solution would indeed be to migrate to `storageNamespace`, which will set all names to the given identifier, giving full seperation across the codebase. However, i'm keen on breaking changes so i want to make it as easy as possible for the user to migrate to this.

The best would be to introduce this change in v11, as v11 will remove the older encryption methods, as well as the migration tool. This way we could also alter `SHARED_PREFERENCES_CONFIG_NAME`, with a new migration tool if needed so users don't lose data.

juliansteenbakker:
I have updated this PR to make the key names configurable. I've added this to a separate PR. You can find my changes here: https://github.com/juliansteenbakker/flutter_secure_storage/pull/1051/changes/b39214d938fdcdafd529090edc7b92a72f7d7c41

What do you guys think, is the problem tackled now with the addition of custom key names?

westracer:
> I have updated this PR to make the key names configurable. I've added this to a separate PR. You can find my changes here: [b39214d](https://github.com/juliansteenbakker/flutter_secure_storage/commit/b39214d938fdcdafd529090edc7b92a72f7d7c41)
> 
> What do you guys think, is the problem tackled now with the addition of custom key names?

LGTM. The only potential issue I see is that instances are created once and cached in the `storagesBySharedPreferencesName` map, but what if any of the storage options change at runtime (e.g., after merging some additional options from a method call, or by creating another storage instance with the same namespace) - is it handled somehow or is it just going to use the first options that were provided?

juliansteenbakker:
> > I have updated this PR to make the key names configurable. I've added this to a separate PR. You can find my changes here: [b39214d](https://github.com/juliansteenbakker/flutter_secure_storage/commit/b39214d938fdcdafd529090edc7b92a72f7d7c41)
> > What do you guys think, is the problem tackled now with the addition of custom key names?
> 
> LGTM. The only potential issue I see is that instances are created once and cached in the `storagesBySharedPreferencesName` map, but what if any of the storage options change at runtime (e.g., after merging some additional options from a method call, or by creating another storage instance with the same namespace) - is it handled somehow or is it just going to use the first options that were provided?

I've fixed this by only setting the config after the initialization check:

```java
    protected void initialize(FlutterSecureStorageConfig config, SecurePreferencesCallback<Void> callback) {
        // Removed here: this.config = config;
        if (preferences != null) {
            callback.onSuccess(null);
            return;
        }
        this.config = config;
```

juliansteenbakker:
I'm closing this in favour of https://github.com/juliansteenbakker/flutter_secure_storage/pull/1051

PR #1036, opened 2026-01-22
Android: Restore native access to FlutterSecureStorage 

Milad-Akarie:
This PR restores access to FlutterSecureStorage from native Android code.
After the latest update, FlutterSecureStorage is no longer usable on Android from the native side, while iOS remains unaffected.

Use case
The Android app includes an interactive home widget. User actions in the widget trigger network calls that require an auth token stored via FlutterSecureStorage on the Flutter side.

What this PR does
It allows creating and initializing a FlutterSecureStorage instance directly in native Android code and enabling secure token access.


ethicnology:
@Milad-Akarie do you think your PR can solve https://github.com/juliansteenbakker/flutter_secure_storage/issues/1043 ?

i5hi:
> @Milad-Akarie do you think your PR can solve #1043 ?

I don't think it does. As it stands it just makes certain classes public so it can be used from native android code

juliansteenbakker:
Thank you for the changes!

PR #1038, opened 2026-01-23
fix: api to 23 again

frananleo:
Downgrade Android minSdkVersion from API 24 to API 23

Summary

This PR updates the Android minimum supported API level from 24 (Android 7.0) down to 23 (Android 6.0) to extend compatibility to older devices.

Changes

Updated minSdkVersion from 24 → 23


frananleo:
Hi! Thanks for the effort to maintain this repo!

I think that in the example app is not needed.

Repo is targeting 3.19 flutter version and the minsdk 24 was introduced later.

Thanks!

juliansteenbakker:
It's indeed not needed, but required for the integration tests. I mean, if we support plugin on sdk 23, we should also be able to test functionality on that version 😄

PR #1041, opened 2026-01-27
Fix #1038: Android minSdkVersion should be 23, not 24

Earbaj:
### Summary
This PR fixes a mismatch between the documented Android minimum SDK version and the actual enforced value.

### Problem
- Changelog states minimum Android SDK is 23
- Package enforces minSdkVersion = 24
- This prevents usage on API 23 devices/projects

### Solution
- Updated package configuration to set minSdkVersion to 23

### Result
- Documentation and implementation are now consistent
- Package works correctly on Android API 23

### DIFF
--- a/flutter_secure_storage/android/build.gradle
+++ b/flutter_secure_storage/android/build.gradle
@@ -38,7 +38,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 24
+        minSdkVersion 23
     }
     
 }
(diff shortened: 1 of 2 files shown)


frananleo:
There is a PR already for this.

https://github.com/juliansteenbakker/flutter_secure_storage/pull/1038#issue-3846409273

Earbaj:
@frananleo Thanks for the heads up. Since this is already addressed in PR #1038, I’ll close this discussion and follow up there.

Issue #1056, opened 2026-02-17
Android ignores biometricPromptTitle and biometricPromptSubtitle in v10.0.0 (key mismatch)


tanawat2542:
## Summary
On Android, `biometricPromptTitle` and `biometricPromptSubtitle` from `AndroidOptions` are ignored in `flutter_secure_storage` v10.0.0.

The biometric prompt always shows fallback defaults:
- "Authenticate to access"
- "Use biometrics or device credentials"

## Environment
- flutter_secure_storage: 10.0.0
- Platform: Android

## Reproduction
1. Configure secure storage with custom prompt text:
   ```dart
   const FlutterSecureStorage(
     aOptions: AndroidOptions.biometric(
       enforceBiometrics: true,
       biometricPromptTitle: 'My Custom Title',
       biometricPromptSubtitle: 'My Custom Subtitle',
     ),
   );

juliansteenbakker:
This issue has been fixed in #1057

PR #1057, opened 2026-02-17
fix(android): align biometric prompt option keys

tanawat2542:
### Summary
- Align Android config keys to `biometricPromptTitle` and `biometricPromptSubtitle` to match the platform options API.
- Keep legacy keys (`prefOptionBiometricPromptTitle` / `prefOptionBiometricPromptSubtitle`) as fallbacks for backward compatibility.
- Refactor string option parsing to reuse shared helper logic.

### Why
The Android key names should match the public options API so values are read consistently.  
This change preserves existing integrations by continuing to support legacy keys.

### Scope
Android-only change with no intended breaking behavior for existing users.

Refs #1056 

Co-authored-by: [napatr-sawa4](https://github.com/napatr-sawa4)


juliansteenbakker:
Thank you!

Issue #1059, opened 2026-02-18
[Android][v10.0.0] Crash: IllegalStateException: Reply already submitted after failed biometric retry "AndroidOptions.biometric(enforceBiometrics: true)"


tanawat2542:
## Summary
On Android with `flutter_secure_storage` v10.0.0, using biometric-enforced options can crash the app when:
1) biometric auth fails once, then
2) user retries in the same prompt.

The app crashes with:
`java.lang.IllegalStateException: Reply already submitted`

## Environment
- flutter_secure_storage: 10.0.0
- Flutter: 3.29.2
- Dart: 3.7.x
- Platform: Android
- Device observed: Xiaomi 2201117SG (MIUI / Android 13+)
- Config:
```dart
const FlutterSecureStorage(
  aOptions: AndroidOptions.biometric(
    enforceBiometrics: true,
    biometricPromptTitle: 'Authenticate to access',
    biometricPromptSubtitle:
        'Use biometrics or device credentials (Device PIN/Pattern/Password)',
  ),
);

juliansteenbakker:
This issue has been fixed in #1083

PR #1060, opened 2026-02-18
fix(android): prevent duplicate biometric result callbacks

tanawat2542:
### Summary
- Keep the biometric prompt active on `onAuthenticationFailed()` (do not complete the operation there).
- Add a one-shot guard in `MethodResultWrapper` using `AtomicBoolean` so only the first callback is delivered.
- Ignore duplicate `success/error/notImplemented` callbacks to prevent duplicate Dart replies.

### Why
On Android, `onAuthenticationFailed()` is non-terminal and may be followed by `onAuthenticationSucceeded()` in the same prompt session.  
Returning an error too early can cause a second result later, leading to:

`java.lang.IllegalStateException: Reply already submitted`

### Scope
Android-only internal behavior fix. No public API changes.

### Validation
- Reproduced flow: fail biometric once, then retry in same prompt.
- Confirmed no duplicate result delivery.
- Ran `flutter analyze` (only existing unrelated lint warning).

Refs #1059


juliansteenbakker:
Thank you for the fix. However, it seems like we don't need the callback function at all. So i will fix it in #1083.

PR #1061, opened 2026-02-19
fix(web): defensive localStorage iteration in readAll() to ignore/remove corrupted entries

eddiejdi:
Problem: iterating browser Storage may throw or return null values for keys (edge-case corrupted Storage), causing runtime errors (e.g. 'Null check operator used on a null value').\n\nFix: make  defensive — safe index-based iteration, null checks for , remove corrupted keys when found, and do not let storage iteration throw.\n\nAlso: add guidance for maintainers to add a unit/integration test to simulate corrupted storage (test suggestion included). This prevents causes of WebView crashes in consumers such as the Estou Aqui app.\n\nChanges: updated web implementation at .\n\nSuggested follow-ups: add a browser integration test that simulates a storage object with a key whose value is null.\n


juliansteenbakker:
Thank you for the contribution. However, an improved check will be merged in #1087

PR #1062, opened 2026-02-19
fix(web): defensive localStorage iteration in readAll()

eddiejdi:
Make readAll() resilient against corrupted Storage entries (null getItem / Storage.key errors). Removes corrupted keys and returns partial results instead of throwing. Includes recommended test suggestion in PR body.


juliansteenbakker:
Thank you for the contribution. However, an improved check will be merged in #1087

PR #1080, opened 2026-03-24
fix(android): guard MethodRunner payload to prevent NPE crash

piaotairen:
## Summary
- Add null/type guards in `FlutterSecureStoragePlugin.MethodRunner.run()` before reading `call.arguments` and `options`.
- Fallback to an empty options map when `options` is missing or not a map.
- Wrap method runner setup in outer `try/catch` so unexpected exceptions are reported through `result.error` instead of crashing the process.

## Why
In `v10.0.0`, malformed/null MethodChannel payloads can trigger `NullPointerException` at `MethodRunner.run`, causing process-level crashes:

```
Exception java.lang.NullPointerException:
  at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin$MethodRunner.run
```

This patch makes the plugin fail gracefully and return an error to Dart.

## Test plan
- [x] Verified diff scope is one file: `flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java`
- [x] Confirmed branch is based on `v10.0.0`
- [ ] Maintainer CI

Made with [Cursor](https://cursor.com)


juliansteenbakker:
@piaotairen there have been some updates on our develop branch. Could you update your accordingly? Thanks!

Issue #888, opened 2025-02-14
MSVC Compilation Error with /permissive- Flag


NEIL-smtg:
Hi, I work on Microsoft Visual C++ testing, After our maintainer acknowledged https://github.com/firebase/flutterfire/issues/16536 and it was being fixed in https://github.com/firebase/flutterfire/pull/16840, we proceeded with further testing and building with `/permissive-` and `/std:c++latest` and encountered several compilation errors.

Note: Feel free to close this issue if /permissive- or /std:c++latest are not supported.

When compiling with /permissive-, we encountered the 2 errors:
In firebase_auth
```c++
flutter_secure_storage_windows_plugin.cpp(55,67): error C4596: 'GetValueKey': illegal qualified name in member declaration

flutter_secure_storage_windows_plugin.cpp(779,31): error C2440: 'initializing': cannot convert from 'ATL::CW2A' to 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' [C:\gitP\firebase\flutterfire\packages\firebase_auth\firebase_auth\example\build\windows\x64\plugins\flutter_secure_storage_windows\flutter_secure_storage_windows_plugin.vcxproj]
```

The source code can be found here: https://github.com/juliansteenbakker/flutter_secure_storage/blob/56d75d5640684f0842eb74e4580f0bd225cdca50/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp#L55

and 

https://github.com/juliansteenbakker/flutter_secure_storage/blob/56d75d5640684f0842eb74e4580f0bd225cdca50/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp#L779

Simply changing the line to:
```c++
std::optional<std::string> GetValueKey(const flutter::EncodableMap *args);
```
and

```c++
std::string target_name = std::string(CW2A(pcred->TargetName));
```

and the error will be resolved.

Repro step:

1. Open x64 native tools command prompt for VS2022
2. set VSCMD_SKIP_SENDTELEMETRY=1 & "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat" -host_arch=amd64 -arch=amd64 & set \_CL\_= /permissive-
3. git clone https://github.com/firebase/flutterfire
4. cd firebase\flutterfire\packages\firebase_auth\firebase_auth\example
5. flutter pub get 2>&1
[...]

PR #1085, opened 2026-04-04
Fix compile on VS2022

DynamicLoader:
Hi，Maintainers.

I would like to fix this per #888, since I am building it on Windows 11 with VS2022. MS has said the STL behavior now have been th same as gcc/clang, so i guess it's better to keep track with latest behavior and ensure the library can be built with modern toolchains. If theres any concerns please let me know, thanks.

### DIFF
--- a/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp
+++ b/flutter_secure_storage_windows/windows/flutter_secure_storage_windows_plugin.cpp
@@ -52,7 +52,7 @@ namespace
         const flutter::EncodableMap *args);
 
     // Derive the key for a value given a method argument map.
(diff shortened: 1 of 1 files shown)


juliansteenbakker:
Thank you for your fix! I will be writing windows tests in the coming week. When done, i will add the tests to this PR and merge if all succeed.

DynamicLoader:
Friendly ping.

Julian Steenbakker ***@***.***> 于2026年4月6日周一 04:02写道：

DynamicLoader:
Hi, I noticed that you just publish new versions with old deps. We have update it in this PR but it seems that the windows component is still 4.1.0. Would you publish it a new version ASAP? Thanks.

juliansteenbakker:
@DynamicLoader please see https://github.com/juliansteenbakker/flutter_secure_storage/issues/1095#issuecomment-4425589249 for more information

PR #1097, opened 2026-04-19
fix: align android default to the flutter one

bakilli:
Flutter default options use KeyCipherAlgorithm.RSA_ECB_OAEPwithSHA_256andMGF1Padding
While android was using DEFAULT_KEY_CIPHER_ALGORITHM = "RSA_ECB_PKCS1Padding"
This meant trying to access keys from native android with default options would cause a migration and broke it.

This simple change updates the Android side default to match the flutter one so people can use secure storage from both sides without specifying an option.

### DIFF
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfig.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfig.java
@@ -19,7 +19,7 @@ public class FlutterSecureStorageConfig {
     private static final String DEFAULT_BIOMETRIC_PROMPT_TITLE = "Authenticate to access";
     private static final String DEFAULT_BIOMETRIC_PROMPT_SUBTITLE = "Use biometrics or device credentials";
     private static final String DEFAULT_STORAGE_CIPHER_ALGORITHM = "AES_GCM_NoPadding";
-    private static final String DEFAULT_KEY_CIPHER_ALGORITHM = "RSA_ECB_PKCS1Padding";
+    private static final String DEFAULT_KEY_CIPHER_ALGORITHM = "RSA_ECB_OAEPwithSHA_256andMGF1Padding";
 
     public static final String PREF_OPTION_NAME = "sharedPreferencesName";
     public static final String PREF_OPTION_PREFIX = "preferencesKeyPrefix";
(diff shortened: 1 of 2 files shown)


bakilli:
https://github.com/juliansteenbakker/flutter_secure_storage/blob/9b0ccf4693a23e46dc484265ec344dbb6a9d4b3d/flutter_secure_storage/android/src/test/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfigTest.java#L81

I forgot to update test case for default key cipher. Updated test case so its aligned as well

juliansteenbakker:
> https://github.com/juliansteenbakker/flutter_secure_storage/blob/9b0ccf4693a23e46dc484265ec344dbb6a9d4b3d/flutter_secure_storage/android/src/test/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfigTest.java#L81
> 
> I forgot to update test case for default key cipher. Updated test case so its aligned as well

Thanks for the update! Will merge after green checks!

bakilli:
> > https://github.com/juliansteenbakker/flutter_secure_storage/blob/9b0ccf4693a23e46dc484265ec344dbb6a9d4b3d/flutter_secure_storage/android/src/test/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfigTest.java#L81
> > 
> > I forgot to update test case for default key cipher. Updated test case so its aligned as well
> 
> Thanks for the update! Will merge after green checks!

I don't know if this is the place for it but you will continue supporting accessing keys from both native and flutter side right? i am currently using different solutions for both of them but i want to combine them in a singular place.

juliansteenbakker:
> > > https://github.com/juliansteenbakker/flutter_secure_storage/blob/9b0ccf4693a23e46dc484265ec344dbb6a9d4b3d/flutter_secure_storage/android/src/test/java/com/it_nomads/fluttersecurestorage/FlutterSecureStorageConfigTest.java#L81
> > > 
> > > I forgot to update test case for default key cipher. Updated test case so its aligned as well
> > 
> > 
> > Thanks for the update! Will merge after green checks!
> 
> I don't know if this is the place for it but you will continue supporting accessing keys from both native and flutter side right? i am currently using different solutions for both of them but i want to combine them in a singular place.

Yes, the native side should be usable without flutter, however right now we have it scoped to only be used by the plugin for security reasons (no public methods). Would you rather want public facing functions? I have to check what for security impact that could have.

bakilli:
I don’t think I’ll need the public-facing functions. I tested reading the key from both sides, and it worked fine. I did not do extensive testing but i don't think it will be a problem

I haven’t tried registering a listener on the native Android side, but I assume it’s possible to listen for changes there and then read the updated value using the plugin’s read function.

One potential issue I noticed—though I didn’t dig into the code—is that if the value is changed on the native side and you try to listen for it on the Flutter side, it might not propagate correctly. It seemed like the listener was more local in scope. That said, I won’t be using it this way, so it’s not a concern for me.

Thanks for your work on this plugin.

PR #1118, opened 2026-05-06
feat: migration tool for secure enclave and resetOnError function

juliansteenbakker:
Adds a migration path for secure enclave


juliansteenbakker:
I've decided to close this for now due to time constraints, but can be opened later.

PR #1123, opened 2026-05-10
Raise Darwin minimums for CryptoKit APIs

Tinnci:
## Draft status

This PR aligns the Darwin package metadata with the CryptoKit APIs used by `flutter_secure_storage_darwin`. The implementation imports CryptoKit and uses APIs such as `SymmetricKey` / `AES`, so the declared minimums need to cover CryptoKit availability.

This is not a metadata-only warning suppression: it keeps the CocoaPods and Swift Package Manager paths consistent by updating both the podspec and `Package.swift`.

## Scope

- iOS minimum: 12.0 -> 13.0
- macOS minimum: 10.14 -> 10.15
- Updates both `flutter_secure_storage_darwin.podspec` and `Package.swift`
- No Dart APIs or runtime code are changed.

## Verification

- Consuming app `flutter build ios --simulator --debug` passed after pinning this fork.
- `git diff --check`


juliansteenbakker:
Thank you for the PR! This is already fixed, and will be released later today.

PR #1132, opened 2026-05-12
fix: linux locked keyring

juliansteenbakker:
Fixes for keyring daemon state and locked keyring state

PR #1141, opened 2026-05-19
docs(readme): improve Linux platform setup / build (libsecret)

EchoEllet:
- Adds Flathub / Flatpak instructions. libsecret is already included in newer runtimes and works out of the box ([Related issue](https://github.com/flathub/shared-modules/issues/424))
- Adds instructions for dnf and pacman package managers
- Separates snapcraft / Ubuntu setup from Flathub and other distros

This makes it easier for developers and users to support and test this plugin on Linux desktops:

- Whether they are an Arch Linux user trying to build the app from source
- Or they want to publish to Flathub
  - Some plugins do not use the supported portals by default (e.g., https://github.com/fluttercommunity/plus_plugins/issues/1241), which results in Flathub rejection. This README change indicates that this plugin complies with Flathub policy.

### DIFF
--- a/flutter_secure_storage/README.md
+++ b/flutter_secure_storage/README.md
@@ -341,7 +341,57 @@ You need the C++ ATL libraries installed along with the rest of Visual Studio Bu
 
 ### Linux
 
-You need `libsecret-1-dev` on your machine to build the project, and `libsecret-1-0` to run the application (add it as a dependency after packaging your app). If you using snapcraft to build the project use the following
+You need to install [Libsecret](https://github.com/GNOME/libsecret) pacakges:
+
+1. **development package**: on your machine to build the project
+2. **runtime pacakge**: to run the application (add it as a dependency after packaging your app).
+
+<details>
+	<summary>Apt / Dnf / Pacman</summary>
+
+For Ubuntu / Debian-based distros (e.g., Linux Mint, PopOS):
+
+```shell
+sudo apt install libsecret-1-0 libsecret-1-dev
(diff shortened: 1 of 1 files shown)


EchoEllet:
[This section](https://pub.dev/packages/flutter_secure_storage_linux#installation) should be also updated. `libsecret-1-dev` and `libjsoncpp-dev` are Debain / Ubuntu package names.

juliansteenbakker:
Thanks for the update! I've added the updated linux package readme doc in https://github.com/juliansteenbakker/flutter_secure_storage/pull/1132.

EchoEllet:
Thank you for reviewing the PR!

> I've added the updated linux package readme doc in #1132.

It seems that the updated instructions were applied [only](https://github.com/juliansteenbakker/flutter_secure_storage/commit/781ae7bd2b08496337f14cacc17001000b38296d) for the `README.md` of `flutter_secure_storage_linux` but sometimes consumers may only read the README of the user-facing package.

Maybe we should update the README of `flutter_secure_storage` to provide a link to `flutter_secure_storage_linux`'s README? Currently, it is snap-only instructions.

juliansteenbakker:
@EchoEllet can you check, and see if it's all correct? I've updated your suggestions here https://github.com/juliansteenbakker/flutter_secure_storage/pull/1146

EchoEllet:
> @EchoEllet can you check, and see if it's all correct? I've updated your suggestions here #1146

LGTM. Thanks!

Issue #1125, opened 2026-05-11
[Android] Issues with enforceBiometrics


Sajito:
I'm trying to use the secure storage to store a pin, which must only be accessible through biometrics.

This is how I build the storage currently.

```dart
final storage = const FlutterSecureStorage(
      aOptions: AndroidOptions.biometric(
        enforceBiometrics: true,
        biometricPromptTitle: 'Authenticate pls',
        storageNamespace: 'biometric',
      ),
      iOptions: IOSOptions(
        accessibility: KeychainAccessibility.passcode,
        accessControlFlags: [AccessControlFlag.biometryCurrentSet],
      ),
    );
```

It seems to work, because on first access biometrics is required. But my issue is, that the authentication seems to be cached somewhere, because reading the value again does not require biometrics again. Instead the value is received without issues.
Also `enforceBiometrics` does not really enforce biometrics, because it also allows using the device pin or pattern.

I'd like to be able to do these two things:
1. Require biometrics on every read.
2. Disable using device pin or pattern.

Using version 10.1.0 of `flutter_secure_storage`.

suesitran:
followup: the PR above fixed half the issue: to select restriction biometric type

The following issue is still not fix:
- reading the value again does not require biometrics again

shan-shaji:
Hi, @juliansteenbakker Thank you so much for the nice plugin.

> reading the value again does not require biometrics again

Is this behavior caused by the decrypted application key being kept in memory and reused to decrypt the stored values? Was this an intentional design decision

theseedhub:
Can the package give users the option to disable the application key caching in FlutterSecureStorage, and let users manage the application key lifecycle in memory themselves? Currently there is no control of how long the keys are presenting in memory, and it also makes subsequent reads no longer guarded by hardware secure storage.

anggaaryas:
Hmm can we close and re init the storage to trigger biometrics?

PR #1143, opened 2026-05-21
feat: add AndroidBiometricType option to restrict authentication to strong biometrics only

CORDEA:
Hi, thank you for always providing the great package!
I'd like to add an option to restrict authentication on Android. I would appreciate it if you could consider this.

## Motivation

  AndroidOptions.biometric() previously accepted both Class 3 (strong) biometrics and device credentials (PIN, pattern, password) with no way to restrict to biometrics alone. Users who need strict biometric-only
  access — e.g. storing a secret that should not be unlockable with a PIN — had no supported path.

  ## What

  Two new options on AndroidOptions and AndroidOptions.biometric():

  - biometricType: AndroidBiometricType (**default: biometricOrDeviceCredential**)
    - biometricOrDeviceCredential — preserves existing behaviour: strong biometrics or device credentials are accepted
    - strongBiometricOnly — restricts to Class 3 biometrics only; device credentials are explicitly rejected everywhere: BiometricManager.canAuthenticate, BiometricPrompt.setAllowedAuthenticators, and
  KeyGenParameterSpec.setUserAuthenticationParameters. Fully enforced on API 30+; best-effort on API 28–29.
  - biometricPromptNegativeButton (default: "Cancel") — the label for the dismiss button on the biometric prompt. Required when using strongBiometricOnly (no device-credential fallback means no implicit dismiss
  action) or on Android 10 (API level 29) and lower where setAllowedAuthenticators is unavailable and a negative button is always needed.

I think it's related to #1125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ final storage = FlutterSecureStorage(
 | `AndroidBiometricType.biometricOrDeviceCredential` | Class 3 biometrics **or** PIN / pattern / password (default) |
 | `AndroidBiometricType.strongBiometricOnly`  | Class 3 (strong) biometrics only — credentials rejected |
 
-> **Note:** On Android 10 (API level 29) and lower, `setAllowedAuthenticators` is not available and the system may still allow device credentials regardless of this setting.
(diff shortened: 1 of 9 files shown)


juliansteenbakker:
This all looks good to me! Lets get it merged!

PR #1167, opened 2026-06-20
feat: add requireBiometricConfirmation option to AndroidOptions

CORDEA:
Hi! I would appreciate it if you could review this.

## Motivation

Android's BiometricPrompt requires users to press an explicit "Confirm" button after passive biometric authentication (e.g. face recognition) by default. This is a useful safeguard for sensitive contexts, but unnecessary for many use cases. Android provides setConfirmationRequired(false) to skip it and enable a lower-friction implicit authentication flow.

https://developer.android.com/identity/sign-in/biometric-auth#no-explicit-user-action


juliansteenbakker:
Thank you for the feature!

PR #1168, opened 2026-06-21
fix: remove redundant ./ prefix from part directives

CORDEA:
Resolves flutter analyze style warnings about unnecessary leading ./ in relative path references.

Related to https://github.com/juliansteenbakker/flutter_secure_storage/actions/runs/27863908065/job/82464593497.


juliansteenbakker:
Thank you for the fix!

PR #1163, opened 2026-06-14
Fixed KeyringLocked error on Linux. Improved error handling on Linux.

dp2022:
Removed SECRET_SERVICE_LOAD_COLLECTIONS from the warmupKeyring. Display real error messages instead of treat all errors as KeyringLocked only.

Issue #778, opened 2024-08-27
[Linux] Warning "libsecret_error: Failed to unlock the keyring"


PureTryOut:
I have an application that uses flutter_secure_storage to locally store credentials. Although I've ran this application successfully before it's been a while since I launched it and now it fails to unlock the keyring for some reason. I'm using KDE Plasma and it's kwallet is used as the keyring provider, with kwalletmanager I can still see the entry from last time I've used flutter_secure_storage so the keyring itself works just fine. I tried wiping the stored credentials and starting anew but it made no difference.

When reading any entry from the storage the console logs the following:

```
** (<appname>19565): WARNING **: 09:53:45.713: libsecret_error: Failed to unlock the keyring
```

It gives me no explanation on what exactly is failing so I have no way to figure out what's going wrong.

PureTryOut:
When calling `.deleteAll()` it actually even errors out badly:

```
** (<appname>:27723): WARNING **: 15:41:52.952: libsecret_error: \xb2\xfe\u001d!
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FormatException: Unexpected extension byte (at offset 1)
#0      _Utf8Decoder.convertSingle (dart:convert-patch/convert_patch.dart:1760:7)
#1      Utf8Decoder.convert (dart:convert/utf.dart:349:37)
#2      StandardMessageCodec.readValueOfType (package:flutter/src/services/message_codecs.dart:502:29)
#3      StandardMessageCodec.readValue (package:flutter/src/services/message_codecs.dart:477:12)
#4      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:644:47)
#5      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:334:18)
<asynchronous suspension>
#6      FlutterSecureStorage.deleteAll (package:flutter_secure_storage/flutter_secure_storage.dart:285:5)
```

PureTryOut:
This is still very much relevant and still breaks the whole functionality of this package on Linux!

juliansteenbakker:
Sorry for the late response. Can you please check if this still an issue with version 10.0.0-beta.1? If so, I will try to setup a linux environment to check how i can solve this issue.

SuricateTrading:
It still doesn't work in 10.0.0-beta.4.
I have a fresh Ubuntu installation, so it can't be me.

gepbird:
[...]

PR #1177, opened 2026-07-09
fix: handle missing default keyring on linux

skylartaylor:
### Steps to reproduce

1. Start a Secret Service daemon without a `default` collection alias.
2. Call `read`, `readAll`, `containsKey`, `delete`, or `deleteAll`.
3. The call fails with `KeyringLocked`, even though the keyring is not locked.

### What

Treat `secret_collection_for_alias_sync()` returning `nullptr` without a `GError` as a missing default collection. Before treating the service as empty, search for matching items in other collections.

When no matching item exists, reads return empty results and deletes are no-ops without creating a collection. If matching data survives outside the missing default alias, or the search is inconclusive, the operation fails closed with `KeyringLocked` so a write or delete cannot create a duplicate collection or orphan existing data.

The Linux integration job now covers:

- a fresh service without a default collection, including safe first-write failure when a headless session cannot prompt
- a missing default alias with an unrelated secret elsewhere, proving the search ignores and preserves nonmatching data
- an initialized and unlocked default collection
- an existing genuinely locked default collection
- a missing default alias with matching data in another collection

The missing-collection fixtures also verify that neither the default alias nor the complete collection set changes during reads, deletes, or a failed headless write.

This change is independent of #1163, which addresses broader Linux error handling.

Related to #778.

Tested with empty missing-collection tests (6), nonmatching-secret tests (6), existing Linux integration tests (15), locked-keyring tests (6), orphaned-collection tests (6), and Linux native tests (13). Formatting and analysis pass, and the workflow changes add no new `actionlint` findings.

Also tested against current Arch packages with KWallet's Secret Service provider (`kwallet` 6.27.0 and `libsecret` 0.21.7). A real Flutter Linux build passed all five noninteractive missing-default operations and left the missing alias and complete collection set unchanged. The write reached KWallet's interactive collection-creation prompt; accepting that prompt still needs a real desktop smoke test because KWallet's headless prompt-dismiss response is not interoperable with libsecret.


juliansteenbakker:
Thanks for the fix, looks good to me!

PR #1182, opened 2026-07-16
refactor(linux): communicate with Secret Service (org.freedesktop.secrets) directly over D-Bus

### LINKED ISSUE #233: Flutter run on linux cannot link secret library
When using flutter_secure_storage 4.2.0 in my app then the build for platform linux fails.

flutter run -v -d linux --no-sound-null-safety
```
[  +38 ms] [4/9] Linking CXX shared library plugins/flutter_secure_storage/libflutter_secure_storage_plugin.so
[        ] FAILED: plugins/flutter_secure_storage/libflutter_secure_storage_plugin.so 
[   +3 ms] : && /snap/flutter/current/usr/bin/clang++ -fPIC -g  -L/snap/flutter/current/usr/lib/gcc/x86_64-linux-gnu/8 -L/snap/flutter/current/usr/lib/x86_64-linux-gnu -lblkid -lgcrypt
-llzma -lpthread -L/snap/flutter/current/usr/lib/ -shared -Wl,-soname,libflutter_secure_storage_plugin.so -o plugins/flutter_secure_storage/libflutter_secure_storage_plugin.so
plugins/flutter_secure_storage/CMakeFiles/flutter_secure_storage_plugin.dir/flutter_secure_storage_plugin.cc.o
-Wl,-rpath,/home/virtual/AndroidStudioProjects/de-app/linux/flutter/ephemeral /snap/flutter/current/usr/lib/x86_64-linux-gnu/libgtk-3.so
/snap/flutter/current/usr/lib/x86_64-linux-gnu/libgdk-3.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libpangocairo-1.0.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libpango-1.0.so
/snap/flutter/current/usr/lib/x86_64-linux-gnu/libatk-1.0.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libcairo-gobject.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libcairo.so
/snap/flutter/current/usr/lib/x86_64-linux-gnu/libgdk_pixbuf-2.0.so /home/virtual/AndroidStudioProjects/de-app/linux/flutter/ephemeral/libflutter_linux_gtk.so
/snap/flutter/current/usr/lib/x86_64-linux-gnu/libblkid.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/liblzma.so -lpkgcfg_lib_LIBSECRET_secret-1-NOTFOUND
/snap/flutter/current/usr/lib/x86_64-linux-gnu/libgio-2.0.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libgobject-2.0.so /snap/flutter/current/usr/lib/x86_64-linux-gnu/libglib-2.0.so
/usr/lib/x86_64-linux-gnu/libjsoncpp.so && :
[   +2 ms] /snap/flutter/current/usr/bin/ld: cannot find -lpkgcfg_lib_LIBSECRET_secret-1-NOTFOUND
[        ] clang: error: linker command failed with exit code 1 (use -v to see invocation)
[ +126 ms] [5/9] Linking CXX shared library plugins/printing/libprinting_plugin.so
[        ] ninja: build stopped: subcommand failed.
[  +33 ms] Building Linux application... (completed in 12.8s)
[+9717 ms] Exception: Build process failed
```

flutter doctor -v
```
[✓] Flutter (Channel stable, 2.0.5, on Linux, locale en_US.UTF-8)
    • Flutter version 2.0.5 at /home/virtual/snap/flutter/common/flutter
    • Framework revision adc687823a (11 days ago), 2021-04-16 09:40:20 -0700
    • Engine revision b09f014e96
    • Dart version 2.12.3

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /home/virtual/Android/Sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: /opt/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at google-chrome

[✓] Linux toolchain - develop for Linux desktop
    • clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
    • cmake version 3.10.2
    • ninja version 1.8.2
    • pkg-config version 0.29.1

[✓] Android Studio
    • Android Studio at /opt/android-studio
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] Connected device (2 available)
    • Linux (desktop) • linux  • linux-x64      • Linux
    • Chrome (web)    • chrome • web-javascript • Google Chrome 90.0.4430.93

• No issues found!
```

sudo apt install libsecret-1-dev libjsoncpp-dev libsecret-1-0 libjsoncpp1
```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libjsoncpp-dev is already the newest version (1.7.4-3.1ubuntu2).
libjsoncpp1 is already the newest version (1.7.4-3.1ubuntu2).
libsecret-1-0 is already the newest version (0.20.3-0ubuntu1).
libsecret-1-dev is already the newest version (0.20.3-0ubuntu1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

### LINKED ISSUE #778: [Linux] Warning "libsecret_error: Failed to unlock the keyring"

### LINKED ISSUE #1181: [linux] xdg:schema (Libsecret schema name) attribute is being set incorrectly
This Libsecret schema:

```c
#include <libsecret/secret.h>

static const SecretSchema my_schema = {
    "org.example.Example",
    SECRET_SCHEMA_NONE,
    {
        { "service", SECRET_SCHEMA_ATTRIBUTE_STRING },
        { "account", SECRET_SCHEMA_ATTRIBUTE_STRING },
        { NULL, 0 },
    }
};
```

Will be converted to this map in normal usage:

```dart
{
  'xdg:schema': 'org.example.Example',
  'service': ...,
  'account': ...,
}
```

Libsecret sets an attribute named `xdg:schema` that uses the name of `SecretSchema` (`org.example.Example'` in this case).

### Steps to reproduce

- Store any secret on Linux.
- Run `secret-tool search account <APPLICATION_ID>.secureStorage` in terminal (repalce `<APPLICATION_ID>` with application ID, for example `com.it_nomads.flutter_secure_storage`)
- The value of `schema` attribute (the raw value is still `xdg:schema` but `secret-tool` normalizes it). The intended value is `<APPLICATION_ID>/FlutterSecureStorage`

### Pointer lifetime issue (likely)

`SecretSchema.name` is initialized with `label.c_str()`, which points to the
`std::string`'s internal buffer. Later changes to the string can invalidate this
pointer, causing Libsecret to receive an incorrect schema name and produce
unstable `xdg:schema` values.

Relevant lines:

https://github.com/juliansteenbakker/flutter_secure_storage/blob/develop/flutter_secure_storage_linux/linux/include/Secret.hpp#L19

https://github.com/juliansteenbakker/flutter_secure_storage/blob/develop/flutter_secure_storage_linux/linux/flutter_secure_storage_linux_plugin.cc#L199-L202

https://github.com/juliansteenbakker/flutter_secure_storage/blob/develop/flutter_secure_storage_linux/linux/include/Secret.hpp#L16

Fixing this bug directly will change the value of `xdg:schema`, which is a **breaking change**. The previously stored secret item needs to be manually migrated, or else a new item will be created after changing the lookup attributes.

EchoEllet:
> [!NOTE]
> This PR should **not be merged** before completing the checklist first.

Rewrites the Linux implementation to communicate directly with the same default backend that GNOME Libsecret depends on (`org.freedesktop.secrets` D-Bus) in Dart without method channels overhead.

Uses [`freedesktop_secret`](https://pub.dev/packages/freedesktop_secret), which uses the [`package:dbus`](https://pub.dev/packages/dbus) package by Canonical (Dart client implementation). I have read [the Secret Service official specification](https://specifications.freedesktop.org/secret-service/latest/) and reviewed my implementation.

> [!TIP]
> `freedesktop_secret` is intended to be a building block for plugins such as `flutter_secure_storage`, `simple_secure_storage`, and `biometric_storage` (rather than a direct replacement).
>
> I encourage reading the README of `freedesktop_secret` for details, scope, and usage.

### Early adoption

Override `flutter_secure_storage_linux`:

```yaml
dependency_overrides:
  flutter_secure_storage_linux:
    git:
      url: https://github.com/EchoEllet/flutter_secure_storage.git
      ref: feat/linux-rewrite
      path: flutter_secure_storage_linux
```

> [!NOTE]
> There was a pointer lifetime bug (#1181) in the previous Linux implementation that is producing an incorrect yet unique `xdg:schema` value per application. To keep backward compatibility with no migrations, we need to reproduce the same pointer bug in Dart, which does not seem to be possible without method channel overhead and native code.
> This PR fixes that issue by running a data migration.
>
> Under normal circumstances, a data migration is not needed; both GNOME Libsecret and this implementation use the same backend over D-Bus (example https://github.com/Skyost/SimpleSecureStorage/pull/15). 

<details><summary>Details regarding data migration</summary>
<p>

This implementation fixes a schema bug in the previous implementation (#1181), and therefore a data migration is required.

By default, it does not delete the old secrets even after migration since it can be destructive and needs to be explicit.

To override this behavior:

```dart
import 'package:flutter_secure_storage_linux/flutter_secure_storage_linux.dart';
import 'package:flutter_secure_storage_platform_interface/flutter_secure_storage_platform_interface.dart';

void main() {
  final instance = FlutterSecureStoragePlatform.instance;
  if (instance is FlutterSecureStorageLinux) {
    instance.deleteLegacyData = true;
  }
}
```

</p>
</details> 

### Justification

I have encountered some issues when using GNOME libsecret in a Flutter app (via method channels), regardless of the plugin or approach I use, and many of them are also commonly reproduced by other developers.
So I wrote [`freedesktop_secret`](https://pub.dev/packages/freedesktop_secret) to address these limitations from the core using Dart.

I have mentioned some of these issues in this PR, but I believe a separate issue is better suited for this topic. Let me know whether you would prefer a GitHub issue, and I will open one with detailed references (and links to other Flutter developers who experienced many similar issues).

I already prepared a detailed Markdown document before creating this PR, but it still needs some polishing and rewriting.
I would appreciate knowing whether you would consider this PR.

### Integration tests

All existing integration tests passed, backward compatibility fully retained, fixes a historical schema bug from the previous implementation (#1181), and automates migration in a non-destructive (does not remove old secret item) way with an option to opt in.

[LinuxDesktopIntegrationTests.webm](https://github.com/user-attachments/assets/cce2e02d-560f-4142-ab91-033a37526ac5)

<details><summary>Tested on KDE, GNOME, Cinnamon and COSMIC</summary>
<p>

[Screencast From 2026-07-19 00-42-34.webm](https://github.com/user-attachments/assets/9c4879ca-a48d-4a33-a680-5529ac01e0d9)

https://github.com/user-attachments/assets/977b4fb5-25f8-4b36-92f6-1369bb5fb329

https://github.com/user-attachments/assets/91db7f93-57b3-4549-91dd-362aae7abddb

</p>
</details> 

`freedesktop_secret` also has [its own internal integration tests](https://github.com/EchoEllet/dart-packages/tree/main/packages/linux_secret_store/freedesktop_secret/integration_test).

## Handles unlock and prompting

Handles lock checking, unlock, prompt, prompt's `dismissed` and `result` (operation-specific) with minimal or no assumptions.

image_2026-07-16_15-25-56

For example, some service implementations will unlock all items when unlocking the default collection. This client does not assume that the item is unlocked or service specific behavior. It tries to do what the specification says clients should handle results without assuming it is unlocked for whatever reason.

### Handles fresh installations

In a fresh installation, the default collection typically does not exist yet, unless the user opens another app that creates it, and then opens the Flutter app. `freedesktop_secret` creates the default collection automatically when calling `storeSecret`.

The previous Linux implementation does not handle this yet, and there is an open pull request to address the same issue #1177 (but with integration tests that verify this behavior explicitly in CI).

### Independent of GNOME Libsecret

The libsecret Linux packages are no longer required to be installed with this implementation by either developers or end users. 

This change makes packaging and publishing the app simpler and more straightforward (at least for Dart/Flutter apps).

No need to install [additional Linux dependencies](https://pub.dev/packages/flutter_secure_storage_linux#installation) (`libsecret`) to build and/or run. The copy of `libjsoncpp-dev` is obsolete now, since this implementation uses `dart:convert` from the Dart standard library.

The user now only needs an active service implementation that complies with the specification (typically already satisfied by default on most Linux desktop environments). This is an existing requirement, not a new one caused by the migration to `freedesktop_secret`.

This also means we **no longer** depend on the system package manager. The package manager approach benefits from distribution repository security fixes, but it comes with tradeoffs: it requires different installation steps (Flatpak, Snaps, AUR, Arch Linux, Ubuntu/Debain, Fedora/Redhat, NixOS, etc.), varies depending on how the application is bundled, and distributions ship different versions, some outdated, some more bleeding edge, and some with custom downstream patches.

The new approach aims to avoid hard-to-reproduce build failures caused by distribution-specific differences, such as [flutter_secure_storage#920](https://github.com/juliansteenbakker/flutter_secure_storage/issues/920) (Ubuntu 25.04+).

<details><summary>Why is this usefl?</summary>
<p>

When a [merge request](https://gitlab.gnome.org/GNOME/libsecret/-/merge_requests/175) gets merged, we would need to wait for distros and Flatpak runtimes to update, and ask users to update for a bug to be fixed.

With `freedesktop_secret`, developers can add temporary `dependency_overrides` (common), patch it via a fork and release an updated version, without asking users questions such as `"What is your distribution, or are you using LTS 24.04?"` or explaining `"PopOS GNOME 22.04 is outdated"`.

Many Linux Users don't want to install a forked dependency that comes with 20 transitive dependencies for running an app.

Example comment on a real app that uses `flutter_secure_storage`: https://github.com/saber-notes/saber/issues/826#issuecomment-1704452332

</p>
</details> 

### Removes legacy workarounds

This [warmup](https://github.com/juliansteenbakker/flutter_secure_storage/blob/develop/flutter_secure_storage_linux/linux/include/Secret.hpp#L98-L142) is no longer needed.

This test secret item is no longer created:

```dart
{'explanation': 'Because of quirks in the gnome libsecret API, flutter_secret_storage needs to store a dummy entry to guarantee that this keyring was properly unlocked. More details at http://crbug.com/660005.'}
```

### Related issues

This fixes existing issues:

- Fixes #1177
  - Note: the PR #1177 contains integration tests and CI setup that should be kept even if the rewrite PR is merged.
- Closes #1163
  - Error handling is already robust and handles some extremely rare cases that never happen in practice. It throws exceptions instead of low-level native errors that are harder to debug, and in some cases, are unclear or cause crashes/freezes.
    - https://github.com/saber-notes/saber/issues/826#issuecomment-1712543740
- Fixes #1181 (and handles migration)
- Fixes #778
- Fixes #233 - [libsecret version mismatch](https://pub.dev/packages/flutter_secure_storage_linux#flutter-installed-via-snap-on-ubuntu-2204) (known issue documented in `flutter_secure_storage_linux`)
  - Since the PR removes the need for `libsecret`, the related section was removed from `README.md` files.

### Updated lookup attributes

The lookup attributes have been changed (with automated migration) to be more flexible. Consider reading [this section](https://pub.dev/packages/freedesktop_secret#for-library-authors-do-not-hardcode-the-application-id).

### Current Limitations of `freedesktop_secret`

They are mostly irrelevant or not very beneficial, but still worth [pointing out](https://pub.dev/packages/freedesktop_secret#limitations).

The first one may read as if the secrets are not encrypted, which is not the actual behavior. Encryption on disk is handled by the service implementation (e.g., KDE Wallet). Consider reading [this section for details](https://pub.dev/packages/freedesktop_secret#supports-only-plain-algorithm-for-transferring-secrets).

<details><summary>What the specification says about D-Bus transfer encyrption</summary>
<p>

> It is strongly recommended that a service implementing this API support the plain algorithm.

> The Secrets API has provision to encrypt secrets while in transit between the service and the client application. The encryption is not envisioned to withstand man in the middle attacks, or other active attacks. It is envisioned to minimize storage of plain text secrets in memory and prevent plain text storage of secrets in a swap file or other caching mechanism.
> 
> Many client applications may choose not to make use of the provisions to encrypt secrets in transit. In fact for applications unable to prevent their own memory from being paged to disk (eg: Java, C# or Python apps), transferring encrypted secrets would be an exercise of questionable value.

Dart implementations generally fall into the same category due to managed memory and garbage collection.

</p>
</details> 

### Secret Portal support

Details #1203
Short version: The updated implementation is focused on `org.freedesktop.secrets`. The previous implementation does not always appear to support the portal in practice. I have created a separate package to address this issue: [flutter_secure_storage_linux_portal](https://github.com/EchoEllet/dart-packages/tree/main/packages/linux_secret_store/flutter_secure_storage_linux_portal). However, it is not published and needs a maintainer review.

### Dart FFI

Dart FFI is used only to retrieve the Linux application ID without Flutter method channels overhead (many Flutter plugins have been rewritten to be FFI-based, e.g., [`path_provider_foundation`](https://github.com/flutter/packages/pull/10722)). And some plugins now use JNI on Android (e.g., [path_provider_android](https://github.com/flutter/packages/pull/9770)).

I have published a package ([linux_application_id](https://pub.dev/packages/linux_application_id)) to minimize maintenance burden. However, removing this dependency is fairly simple ([changes](https://github.com/juliansteenbakker/flutter_secure_storage/compare/78b6e26be9603608e62372614e55da75cbcaa22a..48f77c4d2d47ae1ba791ddc86ad1fa0efd6f80cc)).

### Checklist

- [x] Test [`freedesktop_secret`](https://pub.dev/packages/freedesktop_secret) against different different secret service implementations, distributions and desktops
- [ ] Test it with real published applications
- [x] Implement internal integration tests for `freedesktop_secret`.
- [x] Update `README.md` of both `flutter_secure_storage` and `flutter_secure_storage_linux`
- [ ] Document the change, who it affects, and update the package version.


juliansteenbakker:
Just wanted to say thanks for your work! It looks pretty solid. I do need to run some tests, so i will be testing this during the week. I didn't had much time to work on the plugin last month but coming days i have more free time to test it.

EchoEllet:
Thank you! Feel free to review at any time.

EchoEllet:
@juliansteenbakker I have created a separate issue for the secret portal support (#1203).

I have sent a separate PR (#1204) to fix #1203 (I already have [a portal implementation](https://github.com/EchoEllet/dart-packages/tree/main/packages/linux_secret_store/flutter_secure_storage_linux_portal)).

EchoEllet:
Closing the PR as discussed in https://github.com/juliansteenbakker/flutter_secure_storage/issues/1203#issuecomment-5624556469. I will publish a new package named `flutter_secure_storage_linux_secret_service` in [this](https://github.com/EchoEllet/dart-packages/tree/main/packages/linux_secret_store/flutter_secure_storage_linux_secret_service) repository.

PR #1183, opened 2026-07-22
fix(darwin): restore access to items written before 0.3.0's SecAccess…

### LINKED ISSUE #1158: iOS loses access to stored values after update to flutter_secure_storage 10.1.0
After updating from flutter_secure_storage 10.0.0 to 10.1.0, accessing the secrets stored in the keychain returns nothing and all our iOS users are logged out. When downgrading, the secrets are accessible again. Secrets written with 10.1.0 remain accessible with newer versions. 

This is also reproducible with the flutter_secure_storage example app.
Steps to reproduce:

1. Install example app on iOS device referencing flutter_secure_storage 10.0.0
2. Add some values to the keychain
3. Update dependency to 10.1.0 or newer
4. Stores values are gone
5. Downgrade to 10.0.0
6. Values are back

I'm testing on an iPhone 12 with iOS version 26.5.

juliansteenbakker:
…Control envelope change

Fixes #1158

### DIFF
--- a/flutter_secure_storage/lib/flutter_secure_storage.dart
+++ b/flutter_secure_storage/lib/flutter_secure_storage.dart
@@ -7,13 +7,13 @@ import 'package:flutter/services.dart';
 import 'package:flutter_secure_storage/test/test_flutter_secure_storage_platform.dart';
 import 'package:flutter_secure_storage_platform_interface/flutter_secure_storage_platform_interface.dart';
 
-part './options/android_options.dart';
-part './options/apple_options.dart';
-part './options/linux_options.dart';
-part './options/web_options.dart';
-part './options/windows_options.dart';
-part './options/ios_options.dart';
-part './options/macos_options.dart';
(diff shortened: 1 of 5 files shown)


yanashenyang:
Was this closed intentionally? PR #1183 still shows as open/unmerged and isn't in any release yet, so the fix doesn't seem available to users so far. Mind reopening (or pointing to the release that includes it) so we can track when it actually lands?

yanashenyang:
Closing this as completed is simply wrong. PR #1183 — the fix this issue depends on — is still open, unmerged, sitting on develop, and there is no released version that contains it. Nothing has been completed.

Marking a data-loss bug as "completed" while the fix isn't merged or shipped is misleading: it tells everyone the problem is solved when it isn't, and it hides an issue that's actively logging users out on iOS.

Please reopen this and keep it open until #1183 is actually merged and published in a release. If it was closed by mistake, that's fine — just reopen it.

PR #1209, opened 2026-07-27
fix(android): catch Throwable on worker thread so keystore Errors don't crash the app

### LINKED ISSUE #1208: java.lang.NoSuchFieldError from broken OEM Keystore escapes catch (Exception) and crashes the app (Vivo / Android 12)
## Description

On certain OEM firmware builds (observed on a Vivo device running Android 12), the app hard-crashes on the first `flutter_secure_storage` call. The device's `framework.jar` and its KeyMint HAL classes are mismatched, so **any Android Keystore key generation** throws:

```
Fatal Exception: java.lang.NoSuchFieldError: No instance field tag of type I in class Landroid/hardware/security/keymint/KeyParameter; or its superclasses (declaration of 'android.hardware.security.keymint.KeyParameter' appears in /system/framework/framework.jar!classes2.dex)
       at android.security.keystore2.KeyStore2ParameterUtils.makeBool(KeyStore2ParameterUtils.java:55)
       at android.security.keystore2.KeyStore2ParameterUtils.addUserAuthArgs(KeyStore2ParameterUtils.java:324)
       at android.security.keystore2.AndroidKeyStoreKeyGeneratorSpi.engineInit(AndroidKeyStoreKeyGeneratorSpi.java:260)
       at android.security.keystore2.AndroidKeyStoreKeyGeneratorSpi$AES.engineInit(AndroidKeyStoreKeyGeneratorSpi.java:63)
       at javax.crypto.KeyGenerator.init(KeyGenerator.java:519)
       at javax.crypto.KeyGenerator.init(KeyGenerator.java:502)
       at com.it_nomads.fluttersecurestorage.crypto.MasterKeys.generateKey(MasterKeys.java:97)
       at com.it_nomads.fluttersecurestorage.crypto.MasterKeys.getOrCreate(MasterKeys.java:56)
       at com.it_nomads.fluttersecurestorage.crypto.MasterKey$Builder$Api23Impl.build(MasterKey.java:317)
       at com.it_nomads.fluttersecurestorage.crypto.MasterKey$Builder.build(MasterKey.java:268)
       at com.it_nomads.fluttersecurestorage.FlutterSecureStorage.initializeEncryptedSharedPreferencesManager(FlutterSecureStorage.java:1290)
       at com.it_nomads.fluttersecurestorage.FlutterSecureStorage.initialize(FlutterSecureStorage.java:167)
       at com.it_nomads.fluttersecurestorage.FlutterSecureStoragePlugin$MethodRunner.run(FlutterSecureStoragePlugin.java:168)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:233)
       at android.os.Looper.loop(Looper.java:334)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

On v10.3.1 the trigger is the legacy EncryptedSharedPreferences migration probe: `FlutterSecureStorage.initialize()` calls `initializeEncryptedSharedPreferencesManager()` on a fresh install to check whether v9 data needs migrating, which builds the vendored AndroidX `MasterKey` → `KeyGenerator.init()` → keystore2 → `NoSuchFieldError`.

## Why it is fatal instead of recoverable

`NoSuchFieldError` extends `java.lang.Error`, not `Exception`. Every catch on this path is `catch (Exception e)`:

- `FlutterSecureStoragePlugin.java` — the outer and per-operation catches in `MethodRunner.run()`
- `FlutterSecureStorage.java` — the ESP probe try/catch in `initialize()` and the step-7 ESP check in the backup-protected migration path (v10 only)

So the `Error` propagates out of `MethodRunner.run()`, escapes the plugin's worker `HandlerThread`, and kills the entire process. Nothing reaches the Dart side, so app-level try/catch cannot help, and the fallback logic that already exists (fall through to custom ciphers) never gets a chance to run.

This is not fixed on `develop` / `11.0.0-beta.1` either: the ESP probe is gone there, but the custom cipher implementations (`KeyCipherImplementationRSAOAEP`, `KeyCipherImplementationAES23`) still generate Keystore keys through the same broken framework path (`addUserAuthArgs` runs for every keygen), and the catches are still `Exception`-only — the same devices would crash with a different stack.

## Suggested fix

Widen the `MethodRunner.run()` catches in `FlutterSecureStoragePlugin.java` to `catch (Throwable)` (and `handleException(Throwable)`), so any such failure becomes a `PlatformException` delivered to Dart instead of an app crash. On these devices the keystore itself is broken (typically cleared by a reboot or firmware OTA), so storage may still fail — but apps can then degrade gracefully (e.g. re-prompt login) instead of crash-looping on launch.

PR with this change incoming.

## Environment

- `flutter_secure_storage: 10.3.1` (also affects `develop` by inspection, see above)
- Device: Vivo, Android 12 (API 31)
- No biometric options in use; default `AndroidOptions` with a `preferencesKeyPrefix`
- Reproduction requires affected firmware — the fault is in the device's `framework.jar`/KeyMint pairing, not in any particular plugin configuration

sanjay-aaritya:
## Description

Fixes #1208.

On some OEM firmware builds (observed on Vivo, Android 12), the device's `framework.jar` and KeyMint HAL classes are mismatched, so any Android Keystore key generation throws `java.lang.NoSuchFieldError` from `android.security.keystore2.KeyStore2ParameterUtils`. Because `MethodRunner.run()` only caught `Exception`, this `Error` escaped the plugin's worker `HandlerThread` and killed the whole app process before anything reached the Dart side — app-level try/catch can never see it.

## Changes

- `FlutterSecureStoragePlugin.java`: widen the `MethodRunner.run()` outer catch, the per-operation catch, and the delete-on-failure retry catch from `Exception` to `Throwable`, and change `handleException` to accept `Throwable`.
- `CHANGELOG.md`: add an Unreleased entry.

No behavior change for normal exceptions — they flow through the exact same `handleException` → `result.error(...)` path as before. The only difference is that `Error` subclasses now also surface as a `PlatformException` to Dart instead of crashing the process.

On affected devices the keystore itself is broken (typically cleared by a reboot or firmware OTA), so storage operations may still fail — but apps can now catch the failure and degrade gracefully (e.g. re-prompt login) instead of crash-looping at startup.

## Notes

- The full analysis, including the v10.3.1 stack trace and why the same devices would also crash on `develop` via the custom cipher keygen path (`KeyCipherImplementationRSAOAEP` / `KeyCipherImplementationAES23` → `KeyGenerator`/`KeyPairGenerator` → same keystore2 framework path), is in #1208.
- Deliberately kept minimal: this worker-thread backstop alone converts the process crash into a catchable `PlatformException` for every keystore `Error` raised on this thread. Happy to also widen inner catches (e.g. around cipher initialization) if you'd prefer failures to fall through to internal fallback logic as well.

### DIFF
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
@@ -230,12 +230,12 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
                                 result.notImplemented();
                                 break;
                         }
-                    } catch (Exception e) {
+                    } catch (Throwable e) {
                         if (config.shouldDeleteOnFailure()) {
                             try {
                                 secureStorage.deleteAll();
                                 result.success("Data has been reset");
-                            } catch (Exception ex) {
+                            } catch (Throwable ex) {
                                 handleException(ex);
                             }
                         } else {
@@ -249,13 +249,17 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
                     handleException(e);
                 }
             });
-            } catch (Exception e) {
+            } catch (Throwable e) {
+                // Catch Throwable, not just Exception: some OEM builds throw
+                // java.lang.Error subclasses (e.g. NoSuchFieldError) from Android
+                // Keystore framework code, and an uncaught Error on this
+                // HandlerThread would crash the entire app process.
                 handleException(e);
             }
         }
 
 
-        private void handleException(Exception e) {
+        private void handleException(Throwable e) {
             StringWriter stringWriter = new StringWriter();
             e.printStackTrace(new PrintWriter(stringWriter));
             // Send exception message as the message field so Flutter can parse it
--- a/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
+++ b/flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
@@ -231,6 +231,9 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
                                 break;
                         }
                     } catch (Throwable e) {
+                        if (e instanceof VirtualMachineError) {
+                            throw (VirtualMachineError) e;
+                        }
                         if (config.shouldDeleteOnFailure()) {
                             try {
                                 secureStorage.deleteAll();
@@ -253,7 +256,13 @@ public class FlutterSecureStoragePlugin implements MethodCallHandler, FlutterPlu
                 // Catch Throwable, not just Exception: some OEM builds throw
                 // java.lang.Error subclasses (e.g. NoSuchFieldError) from Android
                 // Keystore framework code, and an uncaught Error on this
-                // HandlerThread would crash the entire app process.
+                // HandlerThread would crash the entire app process. Genuine VM
+                // errors (OOM, StackOverflow) are rethrown instead of being
+                // funneled through handleException/deleteAll, which would
+                // allocate memory the JVM may no longer have.
+                if (e instanceof VirtualMachineError) {
+                    throw (VirtualMachineError) e;
+                }
                 handleException(e);
             }
         }
--- a/flutter_secure_storage/CHANGELOG.md
+++ b/flutter_secure_storage/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Android
+
+- Fixed a fatal app crash when Android Keystore framework code throws `java.lang.Error` subclasses (e.g. `NoSuchFieldError` on some OEM builds with mismatched framework/KeyMint classes): the plugin's worker thread now catches `Throwable` and reports a `PlatformException` to Dart instead of killing the process.
+
 ## 11.0.0-beta.1
 
 **Breaking changes**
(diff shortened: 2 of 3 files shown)


flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -249,13 +252,23 @@ public void onError(Exception e) {
                     handleException(e);
                 }
             });
-            } catch (Exception e) {
+            } catch (Throwable e) {
```
juliansteenbakker, line 255: I've added the change here aswell

flutter_secure_storage/android/src/main/java/com/it_nomads/fluttersecurestorage/FlutterSecureStoragePlugin.java
```diff
@@ -249,13 +252,23 @@ public void onError(Exception e) {
...
+                // Catch Throwable, not just Exception: some OEM builds throw
+                // java.lang.Error subclasses (e.g. NoSuchFieldError) from Android
+                // Keystore framework code, and an uncaught Error on this
+                // HandlerThread would crash the entire app process. Genuine VM
+                // errors (OOM, StackOverflow) are rethrown instead of being
+                // funneled through handleException/deleteAll, which would
+                // allocate memory the JVM may no longer have.
+                if (e instanceof VirtualMachineError) {
```
juliansteenbakker, line 263: I've excluded VM exceptions since the app cannot perform any other actions when such errors are thrown.

juliansteenbakker (approved):
Thanks for the fix!

Issue #1126, opened 2026-05-11
storageNamespace deletes previous data from sharedPreferencesName


iliuta:
Hi,

I upgraded the library from 10.0.0 to 10.1.0
I had an analyze warning because I used sharedPreferencesName which is deprecated.
When I replaced it with storageNamespace and redeployed the app on Android, I lost the saved data.

Going back to sharedPreferencesName did not make the data re-appear.

Here is the code I use:

```
const FlutterSecureStorage(
          aOptions: AndroidOptions(
            storageNamespace: 'FlutterSecureStorageShared',
            preferencesKeyPrefix: 'some-random-constant-string',
            migrateOnAlgorithmChange: true,
            migrateWithBackup: true,
          ),
        );
```

Is there something that I do wrong or there is a specific migration procedure?

iliuta:
Hi @juliansteenbakker ,

I received a notification that you commented on the issue but I no longer see the comment.

I analyzed further the issue and it appears that the old data is not lost but the keys to decrypt it are.

Anyway, I managed to implement a custom migration method, by instantiating a FlutterSecureStorage the old way and another FlutterSecureStorage the new way and then copy all the values from the old instance to the new instance ...

But for me this is a breaking change or even a bug.
If it's a breaking change, this needs to be documented somewhere.

absar:
I think `sharedPreferencesName` should never be deprecated, since if later it's removed, there will be no way for the users to migrate to `storageNamespace`, rather let the user's choose whether they want to use `sharedPreferencesName` or `storageNamespace`. Secondly like the OP said it's a breaking change and should be treated and communicated accordingly by the documentation and change log. Ideally the plugin should provide a flag to auto migrate from `sharedPreferencesName` to `storageNamespace` like how it has done for algorithm change in `v10.0.0`.

absar:
issue still relevant. commenting to avoid getting into stale state

JaySmart001:
Filed #1231 with a failing integration test that demonstrates this issue and traces the implementation of this mechanism from the Android source code. This also occurs in CI. Proposed a solution for the same in the pull request description.

juliansteenbakker:
[...]

PR #1231, opened 2026-08-25
test: add regression test for storageNamespace data loss (#1126)

JaySmart001:
## Description

Adds a failing integration test reproducing #1126, plus what I found tracing it through the Android source on `develop` (v11.0.0).

**The test currently fails by design**, so this is a draft. Happy to mark it skipped, or fold it into a PR with the fix, whichever you prefer.

### Result

Pixel 3a API 34 emulator, run twice against `develop`. The read returns `null` where `'written_before_migration'` was expected.

`+12 ~7 -1` both runs, only this test failing. It doesn't affect the rest of the suite. The namespace in the test deliberately matches the default prefs name, so the data file is identical and the only variable is the key location. It returns null rather than throwing, so there's no signal anything went wrong.

### Mechanism

Four things derive from the config:

| | legacy | with `storageNamespace: 'X'` |
|---|---|---|
| data prefs | `sharedPreferencesName` | `X` |
| config prefs | `FlutterSecureStorageConfiguration:<dataPrefsName>` | follows data prefs |
| wrapped-key prefs | `FlutterSecureKeyStorage` | `FlutterSecureKeyStorage:X` |
| KeyStore alias | `<pkg>.FlutterSecureStoragePluginKeyOAEP` | `<pkg>.FlutterSecureStoragePluginKeyOAEP.X` |

(`getEffectiveDataPrefsName`, `getEffectiveKeyStoragePrefsName`, `getKeyAliasSuffix` in `FlutterSecureStorageConfig`; alias built in `KeyCipherImplementationRSAOAEP:41` and `KeyCipherImplementationAES23:63`; config source constructed at `FlutterSecureStorage:138`.)

When `sharedPreferencesName: 'X'` becomes `storageNamespace: 'X'`, both resolve to the same effective data prefs name, so the data and config markers don't move. Only the wrapped key and the KeyStore alias do. That matches @iliuta's
finding in #1126 that the data survives but the keys don't.

It fails silently because `StorageCipherFactory` treats missing algorithm markers as a fresh install and writes new ones.

### A second problem in v11 (from reading the code, not tested)

v11 removed `sharedPreferencesName` from `AndroidOptions`, but `FlutterSecureStorageConfig:66` still reads it from the options map and falls back to `DEFAULT_PREF_NAME` when absent. For someone who had `sharedPreferencesName: 'X'` on v10:

- Set nothing (the only option now): effective data prefs becomes the default,  so the library reads a different file entirely.
- Set `storageNamespace: 'X'`: the data file is right again, but the key has moved, which is the problem above.

Neither path reaches the data.

### Possible fix

Since the encrypted data never moves in the first case, this may not need a data migration, only the wrapped key rescuing:

1. On init, if the namespaced key prefs are empty but legacy  `FlutterSecureKeyStorage` holds an entry and the unsuffixed alias exists
2. Unwrap the storage key with the legacy alias
3. Re-wrap under the suffixed alias, write to the namespaced prefs
4. Leave the legacy entry so a rollback still works

`NamespacedConfigSource` already does this shape for algorithm markers, namespaced first with a fallback to the legacy global, and `MigrationBackup` plus the `migrateOnAlgorithmChange` flow give a model for the flag, which is what @absar suggested in #1126.

Happy to attempt the migration if that direction looks right. Wanted to check first given the blast radius of getting it wrong here.

Related: #1126

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat: ...`, `fix: ...`, `chore: ...`). This is required by CI.
- [x] Tests were added or updated, if applicable.
- [ ] Documentation (`README.md`) was updated, if applicable.
- [ ] Affected package(s) had their version bumped in `pubspec.yaml` and a changelog entry added to `CHANGELOG.md`.


JaySmart001:
Notable is the CI failure of this new test working as intended: Android Tests (24, default, x86_64) got this failure on CI at API level 24, without any influence from my API level 34 emulator. The other Android build job was aborted due to matrix fail-fast. 21 other checks succeed.

PR #1241, opened 2026-09-07
fix(darwin): skip synchronizable keychain queries when the data protection keychain is off

### LINKED ISSUE #1104: errSecMissingEntitlement (-34018) on macOS when app is signed with Developer ID (outside App Store) — even with sandbox disabled and usesDataProtectionKeychain: false
## Summary

On macOS, calling `FlutterSecureStorage().write(...)` throws `PlatformException(Unexpected security result code, Code: -34018, Message: A required entitlement isn't present., -34018, null)` for an app intended for Developer ID distribution outside the Mac App Store, even when:

- App Sandbox is disabled (both in entitlements XML and `ENABLE_APP_SANDBOX = NO` in the pbxproj)
- `keychain-access-groups` entitlement is not present
- `MacOsOptions(usesDataProtectionKeychain: false)` is explicitly set

The binary's actual signed entitlements (verified via `codesign -d --entitlements -`) contain only `com.apple.security.cs.allow-jit` and `com.apple.security.get-task-allow`. Nothing sandbox-related, nothing keychain-related.

## Environment

- **flutter_secure_storage:** ^10.0.0
- **flutter_secure_storage_darwin:** 0.2.0
- **Flutter:** 3.41.7
- **macOS:** 26.4
- **Xcode:** 26.2
- **Distribution target:** Developer ID Application signing + notarization (no App Store, no provisioning profile)

## Reproduction

Entitlements files (`Runner/DebugProfile.entitlements` and `Runner/Release.entitlements`):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>com.apple.security.cs.allow-jit</key>
    <true/>
</dict>
</plist>
```

`pbxproj` has `ENABLE_APP_SANDBOX = NO` on all build configs.

Dart code:

```dart
final storage = FlutterSecureStorage(
  mOptions: MacOsOptions(
    usesDataProtectionKeychain: false,
  ),
);
await storage.write(key: "Test2", value: "Test");
```

## Actual result

```
PlatformException(Unexpected security result code, Code: -34018, Message: A required entitlement isn't present., -34018, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:653:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:366:18)
<asynchronous suspension>
#2      FlutterSecureStorage.write (package:flutter_secure_storage/flutter_secure_storage.dart:157:7)
```

Verified signed entitlements:

```
$ codesign -d --entitlements - build/macos/Build/Products/Debug/xRay.app
[Dict]
    [Key] com.apple.security.cs.allow-jit
    [Value]
        [Bool] true
    [Key] com.apple.security.get-task-allow
    [Value]
        [Bool] true
```

## Expected

Since the app is not sandboxed, I expected the legacy keychain path (triggered by `usesDataProtectionKeychain: false`) to succeed without any keychain-related entitlement, the same way any non-sandboxed macOS app can call `SecKeychain*` APIs today.

## What I tried

1. **Added `keychain-access-groups` with an explicit group** (`$(AppIdentifierPrefix)dev.xrayinspector.app`) → Xcode demanded a provisioning profile (_"Runner" requires a provisioning profile_). Cannot satisfy this for Developer ID without enabling Keychain Sharing on the App ID + generating a Developer ID profile, which we'd prefer to avoid for distribution simplicity.
2. **Disabled sandbox completely** → build succeeds, but runtime keychain call still fails with `-34018`.
3. **Set `usesDataProtectionKeychain: false`** → no effect; same `-34018`.
4. **Removed the `SystemCapabilities` block from pbxproj** to prevent Xcode re-injecting sandbox → confirmed via `codesign -d --entitlements -` that the sandbox entitlement is not on the binary. Error persists.
5. **`flutter clean` + rebuild** between each attempt.

## Question

What is the intended entitlements configuration for a non-MAS, non-sandboxed, Developer ID signed macOS app using this plugin?

The `flutter_secure_storage_darwin` README suggests adding an empty `<array/>` for `keychain-access-groups`, which I am about to try. If that is indeed the correct answer for Developer ID distribution, it may be worth calling out more prominently in the docs — the current phrasing reads as if it's mainly about App Group sharing, and it's easy to miss that the empty-array form is required even without sharing, and is also the form that doesn't trigger the restricted-entitlement provisioning-profile demand.

If the empty-array form does not resolve it, I'd appreciate any pointers on how to use this plugin in a Developer ID distribution flow without a provisioning profile.

aaronbamblett:
Fixes #1104.

On macOS, an ad-hoc signed app with no `keychain-access-groups` entitlement and `MacOsOptions(usesDataProtectionKeychain: false)` gets `errSecMissingEntitlement` (-34018) from any keychain query carrying `kSecAttrSynchronizable = true`. Two paths in `FlutterSecureStorage.swift` issue that query unconditionally: `performDelete` runs the synchronizable pass before the non-synchronizable one, and `containsKey` runs its synchronizable check first and treats any status other than `errSecSuccess`/`errSecItemNotFound` as failure. `write` calls `containsKey(...).getOrElse(false)`, which swallows the error, so writes look fine while deletes and existence checks fail. Setting `synchronizable: false` in options doesn't help, since `performDelete` overwrites `isSynchronizable` itself regardless of the caller's value.

This adds a private `skipSynchronizableQueries(_:)` helper, true on macOS when `!usesDataProtectionKeychain`, false everywhere else. `performDelete` substitutes `errSecItemNotFound` for the synchronizable pass in that case instead of running the query; `containsKey` skips its synchronizable check the same way. Reasoning: without the data protection keychain there's no synchronizable item to find, so the query has nothing to do even when it's permitted to run. This isn't a behaviour change, it removes a query that could never succeed. Non-macOS platforms aren't touched.

Full disclosure: I don't have Apple hardware, so I haven't built or run this myself. The root cause matches the one described in #1104, and a user of my own app reproduced it and confirmed an equivalent patch resolves login persistence on an ad-hoc signed, sandboxed build with no `keychain-access-groups`.

If you'd rather handle this in `baseQuery` instead of at each call site, happy to rework it.

### DIFF
--- a/flutter_secure_storage_darwin/darwin/flutter_secure_storage_darwin/Sources/flutter_secure_storage_darwin/FlutterSecureStorage.swift
+++ b/flutter_secure_storage_darwin/darwin/flutter_secure_storage_darwin/Sources/flutter_secure_storage_darwin/FlutterSecureStorage.swift
@@ -380,12 +380,15 @@ class FlutterSecureStorage {
             return SecItemCopyMatching(query as CFDictionary, nil)
         }
 
-        // Check synchronizable items first.
-        let statusSync = queryKeychain(withSynchronizable: true)
-        if statusSync == errSecSuccess {
-            return .success(true)
-        } else if statusSync != errSecItemNotFound {
-            return .failure(OSSecError(status: statusSync))
+        // Check synchronizable items first, unless the data protection keychain
+        // is off: see performDelete for why that query cannot run there.
+        if !skipSynchronizableQueries(params) {
+            let statusSync = queryKeychain(withSynchronizable: true)
+            if statusSync == errSecSuccess {
+                return .success(true)
+            } else if statusSync != errSecItemNotFound {
+                return .failure(OSSecError(status: statusSync))
+            }
         }
 
         // Check non-synchronizable items.
@@ -685,6 +688,16 @@ class FlutterSecureStorage {
         return result
     }
 
+    /// True when a `kSecAttrSynchronizable = true` query would be rejected for
+    /// want of an entitlement: macOS, data protection keychain disabled.
+    private func skipSynchronizableQueries(_ params: KeychainQueryParameters) -> Bool {
+        #if os(macOS)
+        return !params.usesDataProtectionKeychain
+        #else
+        return false
+        #endif
+    }
+
     /// Private helper method to perform keychain deletion.
     /// Attempts to delete items with both synchronizable states and without accessibility constraints
     /// to ensure complete removal regardless of how items were originally stored.
@@ -709,7 +722,13 @@ class FlutterSecureStorage {
             return SecItemDelete(query as CFDictionary)
         }
 
-        let statusSync = deleteFromKeychain(withSynchronizable: true)
+        // Without the data protection keychain there is no synchronizable item to
+        // delete, and the query needs an entitlement an ad-hoc signed app cannot
+        // carry: it fails with errSecMissingEntitlement (-34018) and takes the
+        // whole delete with it. See juliansteenbakker/flutter_secure_storage#1104.
+        let statusSync = skipSynchronizableQueries(params)
+            ? errSecItemNotFound
+            : deleteFromKeychain(withSynchronizable: true)
         let statusNonSync = deleteFromKeychain(withSynchronizable: false)
 
         // Return success if both operations report item not found


juliansteenbakker (approved):
Thanks for the fix! Looks good to me.

